PR #57, opened 2015-10-13
Change array order to be NAXIS1 before NAXIS2 etc.

mnorbury:
Whilst using this library I came across a potential bug. The getAxes method in the BasicHDU class reverses the order of the NAXISn dimensions.

With this patch the order is maintained and all unit tests still pass.

Of course, this may be the desired behaviour, but it didn't look right.

### DIFF
--- a/src/main/java/nom/tam/fits/BasicHDU.java
+++ b/src/main/java/nom/tam/fits/BasicHDU.java
@@ -218,7 +218,7 @@ public abstract class BasicHDU<DataClass extends Data> implements FitsElement {
 
         int[] axes = new int[nAxis];
         for (int i = 1; i <= nAxis; i++) {
-            axes[nAxis - i] = this.myHeader.getIntValue(NAXISn.n(i), 0);
+            axes[i - 1] = this.myHeader.getIntValue(NAXISn.n(i), 0);
         }
 
         return axes;


ritchieGitHub:
I think @mnorbury fix is correct, do you concure @TomMcGlynn ? The problem is of course that users could have worked around this bug and by fixing this we break that code... (We could also rename the method to be sure the change is noticed)

ritchieGitHub:
@TomMcGlynn is still offline, sorry.

TomMcGlynn:
Sorry I've been on travel -- currently in Australia.  I've not looked to see if I'm missing some more sublte point, but in FITS files the index represented by NAXIS1 is the index that changes most rapidly.  This reflects the behavior of Fortran where there are true multidimensional arrays.   In Java in a multidimensional array is an array of arrays and the first index is the index that changes slowest.  So at some point a client of the library is going to have to invert the order.  E.g., if I have a FITS file will
    BITPIX=16
    NAXIS1=10
    NAXIS2=20
    NAXIS3=30
this will be read into a Java array 
    short[30][20][10]
so it makes sense to me at least that the returned dimensions are 30,20,10

I think a change in this behavior would break many clients who are relying on the current order.

   Regards,
   Tom McGlynn

ritchieGitHub:
Ok, to make this more clear I have copied your description into the javadoc.

PR #80, opened 2016-02-26
Restored Java 6 compatibility...

attipaci:
The changes restore Java 6 compatibility without affecting functionality. The changes can be summarized in three categories:

1. Filling in diamonds. E.g.:
     ... = new ArrayList<> ();  --->  ... = new ArrayList<HeaderCard>();

2. In util.AsciiFuncs, StandardCharset.US_ASCII is replaced with the local equivalent US_ASCII, which is defined as a private static field.

3. In GZipComressor and GZipCompressor2, the Java 7 resource management try/catch replaced with old-fashioned try/catch, and the resource is explicitly closed whether or not the exception is generated.

That's all. Hopefully it's all good. :-)

-- A.


attipaci:
Almost fully Java 6 back compatible now, except the try-with-resources blocks in GZipCompressor and GZip2Compressor. When changed to a simple try block, it seems that an unexpected IOException is thrown in the block and caught, causing a IllegalStateExceptions in compression the tests.

ritchieGitHub:
When you would use maven, you can test everything  locally ;-) 
don't forget to set the <artifactId>maven-compiler-plugin</artifactId> to 1.6

attipaci:
Thanks for the tip. I'll give it a try. Maven will be new for me also 
(as was github). There is a lot of learning catchup for me to do... I'm 
a pretty pedestrian Java programmer really... Hopefully, I'll get up to 
speed eventually...

-- A.

ritchieGitHub:
@attipaci but please keep in mind that we will only include the 1.6 support if it does not clobber the code to much! But till now it seems to be ok, so keep up the good work. If t becomes to clobbered in die end, I can offer to create a 1.6 branched version.

Important for the debian package would be that it takes the central jar's, and not a own createsd version. Because the central version is pgp-signed and verified.

attipaci:
Hi Ritchie,

I'm in 100% agreement on 1.6. :-)

So far, there was really very little code that required 1.7. Thus, there 
was no strong reason for ditching the back compatibility...

But, at some point there may be desire for more modern code (for example 
we might want to use 1.8 lambda expressions and streams for more 
parallel processing). At that point, I think we should just give up on 
1.6 entirely. I don't have a strong attachment to it, beyond the 
philosophy that if we can provide it easily, then why not... :-)

As for the Debian version, perhaps the best would be if you started a 
discussion with Florian Rothmaier <frothmaier@lsw.uni-heidelberg.de> who 
is the official maintainer of the package. He seems to be a very busy 
guy (which is why the debian package is so outdated). Another good 
contact is Ole Streicher from debian-astro, who is very helpful and 
responsive. You could discuss with them what's the best strategy for 
including nom-tam-fits.

But, I'm afraid it's the distro's strict policy to insist on their own 
build...

-- A.

attipaci:
test suite is now Java 6 compatible also. The majority of code changes are the elimination of try-with-resources in many of the test functions. E.g.:

testSomething() throws Exception {
    try (Fits fits = new Fits()...) {
        // Do stuff with fits, might throw an Exception...
    }
}

These are replaced with:

testSomething() throws Exception {
    Fits fits = new Fits()...
    // Do stuff with fits, might throw an Exception...
    fits.close();
}

The above subtitutions are actually cleaner code, in the sense of eliminating an unnecessary try{} block. When the tests succeed, the resources are all explicitly closed(), leaving no resource leaks (not that we should really care anyway for the tests). Failed tests will have unplugged resource leaks, but then we have bigger problems to worry about...

The other changes are filled diamonds <> with the expected generic types, replacing System.lineseparator() with the equivalent System.getProperty("line.separator") in one instance, and Frame.revalidate() with the equivalent Frame.invalidate() followed by Frame.validate() in another instance.

With these changes, both the libraries and the test suite is Java 6 compatible, and the build succeeds with maven-compiler-plugin source and target set to 1.6...

Fits

ritchieGitHub:
I will check your changes on friyday, thanks for your effort!

attipaci:
The last commit adds a `Fits.write(String fileName)` wrapper method for convenience, which hides the implicit creation, writing to, and closing a suitable BufferedDataOutputStream, and exposes a simple to use method instead to the user. Hopefully we can include this in 1.15 also. (I have been craving this for a long time...) ;-)

ritchieGitHub:
I would be great if you could add unit tests for the new code, -0.8% is a lot.

attipaci:
@ritchieGitHub Sure, I'll look into it. I'm not sure where the -0.8% came from... I only added a bunch of .close() statements, and very few new try/catch blocks. As such, I'm not even sure how these should be tested. I might need some input later...

ritchieGitHub:
maybe this is because of the changes in the master, and coveralls is not "intelligent" enough  to distinguish the branches. In that case no actions necessary. You could pull/merge the master changes into your pull request to check.

attipaci:
Ritchie, I'm afraid you lost me on the last one. I thought I merged the master changes already (which is where the lower coverage must have come from). Is the merge not currently part of the pull request? Do I need to create a new pull request, or do something else? (noob here...)

In any case, even though I'm pretty sure I'm not the reason for the decreased coverage, I don't mind in principle to write some test code for the public good. I might do some of that anyway... :-)

ritchieGitHub:
@attipaci  ok just checked your code, and I now understand the missing test coverage. You worked to "fast" with the tests. Many tests play around with Exceptions and test special cases where exceptions can happen. That means you can not just remove a try block in a test. You have to check if the target of the test is still reached. I have to think how to include you patch.

ritchieGitHub:
I have to include every change manually, the finer details of the tests are just to special. I think tomorrow it will be ready.

attipaci:
Aaah, I see... Before you suffer through it, let me see if I can fix it 
myself (and you can try merge it next week)...

-- A.

attipaci:
I guess, you were a step ahead already... ;-)

-- A.

ritchieGitHub:
No problem, I would have needed to check the changes anyway (because of the test coverage misses). I just hat to take your changes one by one. An do the test stuff the "other" way. Is it ok that I changed your write method. To file parameter, I want to keep the string parameter, unused because of future url detection ( as with read)

attipaci:
Hi Ritchie,

Once again, thanks for taking care of the coverage.

No complaints about String -> File parameter change on write(). I myself 
was torn between the two possibilities. In the end I chose String for 
simplicity, but then again File is more versatile...

perhaps we can still add a String convenience wrapper?:

public void write(String fileName) throws IOException, FitsException {
	write(new File(fileName));
}

-- A.

attipaci:
Ritchie, here is what I propose, in strawman form:

1. write(OutputStream os), and read(InputStream is) methods. E.g.:

```
public void write(OutputStream os) throws IOException, FitsException {
BufferedDataOutputStream stream = new BufferedDataOutputStream(new FileOutputStream(fileName));
}

attipaci:
OK, I think I finally got the previoud comment more or less right (after a few futile attempts...)

ritchieGitHub:
I would say that's a new pull request. I think these methods are good features. So please write them, if you like to do it (including 100% coverage by unit tests). (Attention your StreamLocator class will be a lot more complicated than it seams on first glance ;-) )

The coverage of your unit test's you can find after your local build in:
nom-tam-fits/target/site/jacoco/index.html

attipaci:
Great. That's a plan, then... :+1:

PR #86, opened 2016-03-09
Added the ability to add fixed decimal values to header cards

wcleveland:
I needed the ability to write fixed format decimal to the header cards, and added the following to the API:

In Header:
void addValue(String key, double val, int precision, String comment)

In HeaderCard:
HeaderCard(String key, double value, int precision, String comment)
HeaderCard(String key, float value, int precision, String comment)

In TableHDU:
void setColumnMeta(int index, String key, double value, int precision, String comment, boolean after)

The unit tests for the above classes were modified to cover the added functions.

Github insists on including changes I made in a branch. The changes are in the master branch.

Thanks,
Bill


wcleveland:
The test pass using JDK8 and fail using JDK7. Still working on it.

ritchieGitHub:
Thanks, that looks like a good feature. Could you please make it immutable for the jdk-7 bug? because we support jdk 6/7 and 8. Don't forget to add yourself as a developer in the pom ;-)

wcleveland:
Will do. I'll commit an update this weekend.

ritchieGitHub:
Thaks!

wcleveland:
I changed the method used to create a fixed decimal header card. It takes IEEE 754 into account and rounds the number to the maximum number of significant prior to calling String format.

Checks are made to ensure that the number can be accurately with the specified number of digits. If the value can't an IllegalArgumentException is thrown to indicate that the number of significant digits that would be created is greater than the significant digits stored within the data type.

The maximum significant digits for a double is 16 and for floats is  7.

Looking at the other functions within HeaderCard, I think it may be better to round to the stored significant digits instead of promoting to a BigDecimal. That's an issue for another day.

ritchieGitHub:
Big thanks I will check you PR next Friday. To late for me today, sorry.

ritchieGitHub:
What problem do you see with using a big decimal?

wcleveland:
If the mantissa of the floating point is irrational (in base 2) the number of decimal places returned from the toString may be more than the significant digits stored within source variable. 

The current code base upgrades string representations returned by a double that are greater than 20 needs to BigDecimal. Taking sign and '.' in consideration, that's 18 digits which is still greater than the significant digits that may be stored within double precision floating point.

If the caller expect that many significant digits then maybe the number should already be a BigDecimal prior to the function call.

It's just an early opinion that I haven't fully test yet.

ritchieGitHub:
@wcleveland you pull request is merged, I changed the number formatting to use more standard way's of scaling. and harmonized all "number to string" functions a little. As far as I understood the discussions on bitfits in fits 3.0 the size limitations of numbers are gone, so I also remove the 20 character limitation for numbers.

wcleveland:
Thanks. I won't be able to look at it until I'm "off the clock" this weekend.

attipaci:
See #166 for more on issues relating to base-2 vs base-10 representations, and proposed changes to lessen the confusion...

PR #91, opened 2016-05-18
BinaryTable: Do not encurl variable length columns into multim arrays.

attipaci:
When reading binary tables with variable length columns (e.g. String/byte[] array), the following Exception was thrown:

```
java.lang.RuntimeException: Attempt to curl non-1D array
	at nom.tam.util.ArrayFuncs.curl(ArrayFuncs.java:213)
	at nom.tam.fits.BinaryTable.encurl(BinaryTable.java:951)
	at nom.tam.fits.BinaryTable.getFileRow(BinaryTable.java:1183)
	at nom.tam.fits.BinaryTable.getRow(BinaryTable.java:1313)
	at nom.tam.fits.TableHDU.getRow(TableHDU.java:391)
```

The fix is not to encurl variable length arrays, but to leave them as they are. The fix was tested on a BinaryTableHDU (actual telescope data from APEX) containing variable length columns, and works well.


ritchieGitHub:
Thanks! could you also add a test file to the backbox test repository? Then I will include a blackbox unit test for this.
@TomMcGlynn binaryTable is more your area, could you please verify that the fix is OK for you?

attipaci:
Hi Ritchie,


Of course. I'm travelling until mid next week, with limited internet 
access. But, here is a small test file -- the first 10 rows from an 
actual observatory data file (from APEX, and I think created using 
PyFITS), which has been used and tested with various software packages 
(so it's pretty reasonable to assume that the format is OK).

The last two columns (MONVALUE & MONUNITS) are both variable length 
columns (double[] and byte[], respectively). I also attach a picture of 
the table values shown with 'fv'.

cheers,

-- A.

ritchieGitHub:
@TomMcGlynn could you review this change?

TomMcGlynn:
I've been trying to see where this change is so that I can understand what's going on.  Looking at the old code it looks like curling was already skipped for anything other than strings.so presumably these are variable length string columns.  How do I see what code has been changed and the test case?  I'm sure there is some link which is obvious to those familiar with GIThub but I've been flailing around for a while.   

I've been on travel for the past two weeks but I should be able to look at this (and the boolean issue) now.

ritchieGitHub:
there are two possibilities
- click on the files changed link just under the title of this pull request.
- checkout the code by:
```
git checkout -b attipaci-master master
git pull https://github.com/attipaci/nom-tam-fits.git master
```

attipaci:
Hi Tom,

If you click the link below, you should see the changes directly:

https://github.com/nom-tam-fits/nom-tam-fits/pull/91/commits/7292394c3d352be8c5a7bffce64a903c80ed1b44

The lines highlighted in red were changed to those highlighted in green. 
I.e. after columnToArray(), call encurl() only if it's not a variable 
length column...

Empirically it seems to work well with the variable length columns I 
tried to work with... But perhaps there is another (better) way to 
achieve the same?

cheers,

-- A.

TomMcGlynn:
I'm having trouble matching my version with the changes but presumably 
that's because there are other changes.  If I understood Git and GitHUB 
I'd presumably do better.  That said my sense is that this is the wrong 
place to fix this.  Encurl is intended to be a routine which knows how 
and if to transform the linear version of an array into the requested 
version.  If it's not appropriate to curl a variable length string 
column, then I think the fix goes there.  The fact that you had to do it 
in two places suggests that too.

However I'm not clear why this is occuring.  The row dimensionality 
should only be 1 and encurl doesn't curl strings with dimensionality 
less than two.  Clearly there's an incorrect assumption in the code and 
I'd like to track that down.  Did you put the file that stimulates this 
bug in the repository?

     Tom

Attila Kovacs wrote:

attipaci:
Hi Tom,

Yes, I put 'bt13.fits' under the target/ folder. That's 10 rows of 
binary table with the last two columns being variable. MONVALUE is a 
variable length double[] array, while MONUNITS is a variable length 
byte[] (string) array.

Before I created my fix, I noticed that MONVALUE was returned as 
byte[][] by columnToArray(), and passed to encurl() as such, which 
caused the problem (since, as you say encurl() expects a 1D array...)

I'm not sure if any of it helps...

-- A.

attipaci:
PS.

I also tried changing how columnToArray() deals with byte[] arrays, but 
then I had problems with fixed-length string/byte[] columns -- this is 
why I settled on not calling encurl() for variable-length columns. That 
seemed to work for all cases...

-- A.

ritchieGitHub:
Attention, do not check-in anything under target. Use src/test/resources (and an appropriate subdirectory). The target directory is deleted in the "clean" phase, and should contain only generated files during the build. All files in src/test/resources will be copied to target/test-classes in the phase "process-test-resources", so in a unit test you can access them directly or as part of the classpath (target/test-classes).

attipaci:
Thanks for pointing this out, Ritchie. I put the image as 
'varlen-bintable.fits' under my fork of the blackbox images, and created 
a pull request... I hope that's a better way to go about it. (yup, using 
git does take some getting used to...)

ritchieGitHub:
already merged. @TomMcGlynn the test file is now available from the blackbox images repository.

TomMcGlynn:
OK...  I think the real problem was in the BinaryTable.processCol 
function in the lines:

         if (colDesc.isVarying) {
             dims = new int[]{
//                this.nRow,
                 2
             };

where the commented out line was the culprit.  This is supposed to be 
setting the dimensions for a column entry
in  a single row, so the addition of nRow is wrong.  It doesn't affect 
the processing of variable length rows too much
because the pointer structure is known to the code and the actual length 
is gotten from that.
So the bug persisted for a long, long time.  But it changes the apparent 
dimensionality of variable length strings
from 1 to 2 and that means that encurl tries to curl them.

I haven't checked to see that this actually fixes Attila's issue, but it 
feels more 'right' to me then the ad hoc solution.

It didn't take too long to find this, but I had to spend a long time 
learning about new features that seem very appropriate but I hadn't 
understood.  E.g., when I try to build code NetBeans runs the formatting 
checks so if I try to add a debugging
line and do something like:
    System.err.println("Parameter="+parameter);
it fails the formatting check.  We also appear to be perilously close in 
BinaryTable to a 2000 line limit on classes.  After putting in the 
change one of the regression tests failed.  I originally thought this 
was badCase2 since I could see that 'failing' in the output.  Took a 
while to realize that that was a normal logging output.
The actual failure was in testSet which tests BinaryTable.getDimens 
whose output is changed but only because it was wrong before.  It took a 
while to understand that while NetBeans would run the tests when I did a 
build, it would run them and actually tell me which failed if a did a 
test instead.

Note that there is an error in the way this was tested.  The function 
ArrayFuncs.flatten was used to flatten a non-rectangular array.  That's 
doesn't really work properly (it was intended for flattening box like 
arrays).  e.g.,
     flatten(new int[][]{{1,2,3},{4,5}, {6}});
will return an array with 9 elements or just 3 if the subarrays are 
reversed.

In this case I don't think it was so much GITHub that was the problem 
but the way Maven and NetBeans interact.   When I try to put the changes 
back in -- then I'll  likely have GITHub issues!


     Tom


Attila Kovacs wrote:

attipaci:
Thanks Tom,

Yes, I agree that looks better. :-) I will try it locally myself and see 
if it works...

-- A.

attipaci:
Yes, it works! Thanks.

attipaci:
Yup, the second-to-last test in BinaryTableTest.testSet (lines 1082-1099) fails. (Sorry I cannot test locally, I get a version number error for JaCoCo, and I have not figured out how to work around it...) 

If I understand correctly, Tom's saying that flattening variable length columns does not make too much sense. Perhaps, we should skip this test?

TomMcGlynn:
The test of the getDimens function fails because it is returning the 
wrong results with the extra nRow values.
That test needs to be updated.

     Tom

Attila Kovacs wrote:

attipaci:
Tom, as I am trying to include variable string columns (which wasn't tested before), I noticed another peculiarity. 

If I add Strings of different length to the column, the table created from it has fixed-length String columns (with length equal to the longest entry). However, it does create the proper variable length table with varied byte[] arrays instead of Strings. So, the functionality is there, and is accessible. 

Is that the intended behavior?

attipaci:
I'm happy with it. 'bt2' now includes a variable length byte[] column, which was missing before, and the tests have been adjusted to work with it. Time to pull? 

Ritchie, what's your ETA for 1.15.0?

TomMcGlynn:
Attila Kovacs wrote:
>
> Tom, as I am trying to include variable string columns (which wasn't 
> tested before), I noticed another peculiarity.
>
> If I add Strings of different length to the column, the table created 
> from it has fixed-length String columns (with length equal to the 
> longest entry). However, it does create the proper variable length 
> table with varied byte[] arrays instead of Strings. So, the 
> functionality is there, and is accessible.
>
> Is that the intended behavior?
>
> —
> You are receiving this because you were mentioned.
> Reply to this email directly, view it on GitHub 
> <https://github.com/nom-tam-fits/nom-tam-fits/pull/91#issuecomment-222259394>, 
> or mute the thread 
> <https://github.com/notifications/unsubscribe/ADIB3KmEki-pxLl0fcHrToPyX9jSF0Qxks5qF2TbgaJpZM4IhhgJ>.
>
That's intended.  Historically and still to some extent, FITS' readers 
ability to support variable length arrays has been spotty, so the 
library requires that a user has to indicate that they explicitly want 
the capability.  Since the typical situation with a string column is 
that the strings in different rows would be of different lengths we use 
padding rather than  variable length arrays. My guess is that it's still 
the right choice but the decision is a bit closer now.  The big 
disadvantage to padding is that new rows are limited to the maximum of 
the original set -- but if the original set of strings were all of the 
same length, we'd still face that problem unless we always use variable 
length arrays for scalar string columns.

     Tom

ritchieGitHub:
We can do a 1.14.3 any time, I would propose Tomorrow or next week. The only thing missing is a unit test that tests the blackbox image (read/rewrite/read) .
I will write that test, @attipaci would you check the test when it is ready and correct it if your case needs something specific?

ritchieGitHub:
a test is included:
```
nom.tam.fits.test.UserProvidedTest.testBinaryTableEncurlProblem()
```
 @attipaci please verify that this realy tests your case.

attipaci:
Hi Ritchie,

Thanks for creating that new test. It looks pretty thorough. :-)
I look forward to the release...

-- A.

PR #92, opened 2016-05-28
removal of redundent attribute "rowSize" 

ritchieGitHub:
@TomMcGlynn the rowSize attribute caused problems during the binary compression implementation. And will also cause problems when you add a empty column (no rows yet) and add the rows afterward (fixed that in the head).  

But the rowSize attribute is redundant and not necessary, as far as I can see. This pull request contains the changes necessary to delete the attribute, and still have all tests succeed.

@TomMcGlynn could you verify, that my change is correct? If it is just push the merge pull request button.

### DIFF
--- a/src/main/java/nom/tam/fits/BinaryTable.java
+++ b/src/main/java/nom/tam/fits/BinaryTable.java
@@ -143,6 +143,20 @@ public class BinaryTable extends AbstractTableData {
         }
 
         /**
+         * @return new instance of the array with space for the specified number
+         *         of rows.
+         * @param nRow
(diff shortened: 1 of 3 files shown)


TomMcGlynn:
This looks fine.  However I think that the release notes need to include a mention note that the deletColumns method in ColumnTable has changed from int to void.  I don't think it's too likely that anyone is using that class somewhere else, but it's a breaking change to the nominally public interface.

ritchieGitHub:
Ok, I will write it in the  release notes

PR #101, opened 2016-08-16
Fix introduction

MaxNoe:
All examples are tested to work with the current version.

I removed the package summary, because this is better found in the javadocs.


ritchieGitHub:
Great! big thanks! I will check the changes in the next few days

PR #119, opened 2017-07-07
Added Scientific Notation

wcleveland:
Scientific notation is used when added to the Header with:
hdr.addValue(key, numberValue, fractional_digits, useD, comment). API explained in javadoc.

The library will also read scientific notation that uses 'D' instead or 'E' (e.g. 1.234D+2) as a number value.


wcleveland:
Going to add more Header methods and some more test coverage.

PR #121, opened 2017-07-14
Improved Header card ordering...

attipaci:
Header card ordering was unpredictable when editing existing headers, since the current position, at which new cards were added, would jump around unpredictably whenever an existing key was updated or removed, or when certain Header methods were called that used an iterator. 

This pull request is an effort to give the user control over the header ordering in a transparent way.

Consider the example, of the user wanting to insert a number of header keys before 'MARKER':

```
   h.setKey("MARKER");
   h.addValue("KEY1", ...);
   h.addValue("KEY2", ...);
   h.addValue("KEY3", ...);
   h.addValue("KEY4", ...);
   h.addValue("KEY5", ...);
```

Say, keys KEY2, and KEY4 already existed in the header at some (to us unknown) location. Before this fix, what happened was:

* KEY1 was inserted at the requested location before MARKER.
* KEY2 and KEY3 end up at the 'unknown' prior location of KEY2.
* KEY4 and KEY5 end up at another 'unknown' prior location of KEY4.

What is expected is this:

* KEY1 through KEY5 are added before the MARKER. Prior occurrences of KEY2 and KEY4 are removed. 

The first step towards providing the expected behaviour is to ensure that the current position moves only when we explicitly want it to move, e.g. by the Header.findCard(), Header.nextCard(), or Header.positionAfterIndex() methods. In situ key updates, e.g. via Header.updateLine(), and removals, e.g. via Header.deleteKey(), should not affect the current position unwittingly. Nor should any Header operations, other than the above mentioned methods used explicitly for repositioning...

Secondly, until now removals of existing keys, when inserting or removing entries (internal to HashedList) did not keep the current position updated. Fixing this required a bigger cleanup:

 1. current position is now maintained inside HashedList, instead of Header,
    so consistency can be maintained through all HashedList operations.

 2. New HashedList cursor() and the private Header cursor() methods, now
    return the HashedListIterator object that keeps the current position in the header.

 3. Header operations that use an iterator now use an ad-hoc iterator instance
    instead of messing with the current position.

HeaderOrderTest was updated to test additions (to end), insertions (at specific locations), and updates as well.

-- A.


attipaci:
More stuff is on the way...

wcleveland:
Is there an open issue describing the behavior that you are trying to fix?

ritchieGitHub:
One important thing is, that api functionality should not change. when I read the changes correctly some methods will behave in an other way as before -> api change.
One other issue is that the header api is already very broad (list of many methods). I think we should not add more of them. We should leave that for a 2.x branch, where we can use java 8 features to reduce the methods (2.x may break the api).

attipaci:
No changes to API (no new public or protected methods have been added or removed).

TomMcGlynn:
Richard van Nieuwenhoven wrote:
>
> @ritchieGitHub <https://github.com/ritchiegithub> requested your review on: 
> nom-tam-fits/nom-tam-fits#121 <https://github.com/nom-tam-fits/nom-tam-fits/pull/121> Improved 
> Header card ordering....
>
This suggestion doesn't seem to change the interface at all, but it does change the behavior, so it 
could potentially break someone's existing code.  In practice however, it seems unlikely that a user 
would both simultaneously be writing FITS headers that would get disordered because of preexistence 
of some of the FITS keywords and also require the FITS keywords to be in a specific order.  However 
although it seems unlikely to break real requirements, I could envisage it breaking regression tests 
that simply do byte by byte comparisons even though the code doesn't really care about the order of 
keywords.

Nonetheless I think it's worth doing.  While in principle the order of keywords within a FITS header 
should usually matter, I'd expect that the net gain of having better organized headers will offset 
the rare problems that this might cause.

The release notes will need to have a prominent warning.

     Tom

attipaci:
Hi Tom,


Thanks for weighing in.

The good news is that all the regression tests passed with flying 
colors, even after the change. :-)

The change is unlikely to break anyone's code, even if that is a 
theoretical possibility, which I admit. However, the behaviour remains 
the same for simple header constructs (e.g. additions only). On the 
flipside, anyone that uses complex header edits had to deal with 
semi-random ordering already, as a result of the existing behavior. (It 
is generally not a good idea to rely on header order when parsing 
headers -- and I think most users already know that...).

With the change, headers can be properly organized for human consumption 
also. I often visually inspect headers when trying to identify problems. 
When the headers are long and complex, it's a good idea to break it up 
into sections separated by easily visible comment lines, so you can have 
a good overview of things that belong together. But, when keys end up at 
random places (which is what kept happening), it all becomes a bloody 
nightmare...

I think we are on the same page, as to why this change is a good idea...

In any case, here is the example that I put in the pull comment:

Consider the example of the user wanting to insert a number of header 
keys before 'MARKER':

    h.setKey("MARKER");
    h.addValue("KEY1", ...);
    h.addValue("KEY2", ...);
    h.addValue("KEY3", ...);
    h.addValue("KEY4", ...);
    h.addValue("KEY5", ...);

Say, keys KEY2, and KEY4 already existed in the header at some (to us 
unknown) location. Before this fix, what happened was:

  * KEY1 was inserted at the requested location before MARKER.
  * KEY2 and KEY3 end up at the 'unknown' prior location of KEY2.
  * KEY4 and KEY5 end up at another 'unknown' prior location of KEY4.

What is expected is this:

  * KEY1 through KEY5 are added before the MARKER. Prior occurrences of 
KEY2 and KEY4 are removed.


So, the fixes are aimed at providing the expected behaviour 
*consistently*, and not just when adding new keys...

-- A.

PR #122, opened 2017-07-19
HashedList and HeaderTest leftovers from prior pull request...

attipaci:
Just a minor fix to HashedList.internalRemove() -- an off-by-one index in the conditional check. HeaderOrderTest has been reinstated to near its original state, and instead a new orderTest() was added to HeaderTest. (I may continue to expand that test a little further still...)

### DIFF
--- a/src/main/java/nom/tam/util/HashedList.java
+++ b/src/main/java/nom/tam/util/HashedList.java
@@ -98,16 +98,13 @@ public class HashedList<VALUE extends CursorValue<String>> implements Collection
 
         @Override
         public void add(VALUE reference) {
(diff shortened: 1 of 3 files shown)


ritchieGitHub:
Great I will wait with merging till you say so ;)

attipaci:
Hi Ritchie! OK. I think it's ready for the merge...

ritchieGitHub:
Ok, just a note the code is auto formatted so added blanks in comments won't stay there. If you want a special format use the javadoc techniques.

ritchieGitHub:
just a note, I have sende you a private email

PR #138, opened 2019-07-14
FitsDate equals/hashCode/compareTo

### LINKED ISSUE #136: FitsDate should (at least) override equals
The FitsDate class as it is currently does not allow any kind of access to the parsed components, nor comparison between instances. The only functionalities it supports are:
* creation from a String and casting back to it (through the toString override).
* creation from java.util.Date and casting back to it.

However, FitsDate supports dates with leap seconds, something that is not directly or universally supported by Date/Calendar/Instant. Thus, it would be helpful it FitsDate could be modified to implement value-based equality semantics, overriding `boolean equals(Object)` to compare all data fields (except for date, which is a cached value) and `int hashCode()` with it to respect the API contract. Even better, it could implement `Comparable<T>` and add accessors for the fields.

FinitePhaseSpace:
Implemented equals, hashCode and compareTo
Closes #136


attipaci:
I have incorporated the fixes from this pull request into an unofficial 1.1.5.3-AK2 release on [my fork](https://github.com/attipaci/nom-tam-fits). 

Also, in case you have any further fixes, I'll plan to keep regularly integrating pull requests made to the base repo, at least until the base repo is actively mainained again. (I could no longer bear that so many good fixes were unheeded here...)

PR #142, opened 2020-08-25
MAX_INTEGER_LENGTH is 11 not 10

mbtaylor:
Although the decimal representation of Integer.MAX_VALUE (2^31-1)
has 10 characters, it's 11 for Integer.MIN_VALUE (-2^31).
Prior to this fix, ASCII table columns with format I10 were
interpreted as 32-bit integers, which meant that large negative
values could not be read correctly.
Now I10 are read as 64-bit integers.


mbtaylor:
Wrong!  I got the logic backwards for this commit.  **Do not use.** See instead #143, where I've got it the right way round.

PR #143, opened 2020-08-25
I10-format ASCII table columns should be long, not integer

mbtaylor:
Prior to this fix, ASCII table columns with format I10 were
interpreted as 32-bit integers (on the grounds that Integer.MAX_VALUE
has 10 characters?), but some 10-character numbers cannot be
represented in 32 bits, which meant that some potential values
in 10-character columns (in the range 2147483648-9999999999)
could not be read correctly.  Fix it so that I10 and larger,
not I11 and larger, are treated as 64-bit.


attipaci:
The change breaks a number of test cases. Specifically:

```
Failed tests: 
  AsciiTableTest.test:505->createByColumn:94->checkByColumn:116 ByCol:1 expected:<true> but was:<false>
Tests in error: 
  BaseFitsTest.testFitsDeleteHdu:175 ClassCast class [J cannot be cast to class ...
  BaseFitsTest.testFitsSkipHdu:137 ClassCast class [J cannot be cast to class [I...
```

It seems to me that while the rationale for the change is generally sound, it does unexpectedly change the data container on read-back from `int[]` to `long[]` unexpectedly when writing regular 32-bit integers in the range 1000000000 to 2147483647. This will fail any round-trip tests with ASCII tables that use 32-bit integers in that range...

BTW, I am trying to 'maintain' a fork of nom-tam-fits on my [remix](https://github.com/attipaci/nom-tam-fits/tree/remix) branch, with all pending pull-requests integrated, so I can build packages with the latest fixes. If you are interested on working more on this issue, feel free to create a pull request into my remix branch, and I will integrate it if possible...

mbtaylor:
I added a new commit that makes this behaviour configurable.  By default the previous round-trip behaviour is maintained, so the unit tests will work without modification, but I've introduced an alternative constructor with a flag that lets you choose for I10 to be interpreted as longs. It's a bit messy, but it would solve my problem.
(I actually think that round-trip-preserving is less important than being able to read all values in an I10 column, but changing the default behaviour may be more trouble than it's worth to the unit tests and legacy code, so this commit doesn't do that).  @attipaci what do you think?

attipaci:
OK, I think that is passable. The only thing I don't like is the name `supportRoundTrip`, since that is test specific, but not informative to devs otherwise. Could we call it `prefersInt` (or `prefersLong` if negated) instead? And in the javadoc mention that this determines whether I10 ASCII columns are interpreted as `int` or `long`?
Otherwise, I'm OK with it...

attipaci:
@mbtaylor, Could you also add a test case with I10 columns (either as a blackbox image, or perhaps easier if just an ASCII table you create on-the-fly). Then, you should test with the new `boolean` argument being set both to `true` and to `false`, and with integer values below and above the 32-bit limit (e.g. checking for excpetions or what not when trying to read large i10 value as an `int`)...

attipaci:
As for collapsing commits, I really don't care. In my mind it's historical revision. I think it's more natural to keep the full trail, with mistakes and all. The only thing I hope is that HEAD on master is always a good build...

PR #153, opened 2021-09-01
Fix-up build, no Logger warnings on multiple CONTINUE keywords, and other small fixes.

attipaci:
It overrides issue #152, with just the intended maintenance fixes, such as:

- Bump to Java 8 (Java 6 is a security risk, and 8 is now widely adopted everywhere).
- Update jacoco to latest version to fix build.
- Update findbugs to final version 3.0.5 to fix build.
- Specify javadoc version to 8 to fix javadoc build.
- Change non-standard `<tt>` HTML tags to standard `<code>` tags (to fix javadoc errors).
- No Logger warning for multiple `CONTINUE` keywords (these can appear any number of times, like `COMMENT` and `HISTORY`).
- Remove override of `finalize()` in BufferedFile -- it trips up findbugs, and not a proper way to handle a resource anyway.
- Fix duplicated @return tags
- More tolerant reading of HIERARCH keys to match the more lax writing of them. Also, more consistent processing of non-hierarch cards whether or not hierarch processing is enabled.

-- A.


attipaci:
This pull request is also incorporated into an unofficial 1.1.5.3-AK2 release on [my fork](https://github.com/attipaci/nom-tam-fits).

Issue #148, opened 2020-12-18
Unwanted quote manipulation for string headers


mbtaylor:
The `nom.tam.fits.HeaderCard` constructor for String values includes this code:

    if (value.startsWith("'")) {
        if (value.charAt(value.length() - 1) != '\'') {
            throw new HeaderCardException("Missing end quote in string value");
        }
        value = value.substring(1, value.length() - 1).trim();
    }

I think it's trying to be helpful in case you've pre-quoted a string value, though this behaviour is not documented.  One consequence is that e.g.

    new HeaderCard("TUNIT67", "'electron'.s**-1", "Units for phot_g_mean_flux");

fails with an exception.  The only way you can set a card with that value, if you know the trick, is to wrap it in single quotes:

    new HeaderCard("CARD", "''electron'.s**-1'", "comment");

in which case the code strips out the outer quotes.  This behaviour is surprising and hard to work with.  I'd like to remove the above stanza/behaviour altogether, but I don't know whether other code depends on it.

attipaci:
I tend to agree that that's an over-zealous check. I'd be in favor of its removal, provided it does not break any of our existing test cases. I'll look into it over the next week or two. Feel free to remind me if you don't hear back from me in say 2 weeks, or feel free to preempt me with a pull request... :-)

attipaci:
OK, now that I slept on it a bit, I know why the check is there -- it's because in the FITS convention, the string values must be quoted with single quotes when the header is wrotten. So when you parse the header from a FITS file, you want to remove those quotes. However, if you are creating a `HeaderCard` with a `String` value argument (as in your example), it should be handled separately (with no checks), but right now it's not...

attipaci:
Hi Mark,

I have started on this issue, and I think it is more or less resolved. I want to sleep on it to make sure I am not forgetting anything and that I am testing for various possibilities. I'll put it on a branch on my fork, so you can take a look also. Alternatively, I can just pull my changes into the mainline, and you can then see if you want to do any add-on work after that...

mbtaylor:
@attipaci, thank you for this, it passes my unit tests now.  I haven't tested it in all situations (e.g. long string values) but it looks OK to me, apart from the minor review comment I made on the PR.

PR #156, opened 2021-09-26
Issue 148: Handling single quotes

attipaci:
Issue #148, for handling single quotes in `HeaderCard` created with a `String` value, and corresponding test cases.


src/main/java/nom/tam/fits/HeaderCard.java
```diff
@@ -786,14 +787,18 @@ private HeaderCard(String key, String value, String comment, boolean nullable, b
...
-            value = value.replaceAll(" *$", "");
+            // Discard trailing spaces (TODO do we really want to?)
+            value = value.replaceAll("[ \t\r\n]*$", "");
 
+            /*
             if (value.startsWith("'")) {
                 if (value.charAt(value.length() - 1) != '\'') {
                     throw new HeaderCardException("Missing end quote in string value");
                 }
                 value = value.substring(1, value.length() - 1).trim();
```
mbtaylor, line 799: Just commenting out this code may be confusing to later readers. I would suggest either removing it altogether, or adding a comment about what's gone on here.

mbtaylor (changes requested):

src/main/java/nom/tam/fits/HeaderCard.java
attipaci: Yes, that's necessary because it is used to parse quoted string from the header, i.e.:

```
 MYKEY   = 'some string value ' / Comment here
```
However, this method should not be called when constructing a card with a `HeaderCard()` constructor with a `String` value given programmatically. And that is the change I made (among other things..)

I also added test cases for quoted values in the string value itself. So I think we are good. 

However, I am still working on the parsing of header lines, because there is a lot of wonkiness there, and it's more trouble waiting to happen...

mbtaylor: I'm not objecting to the revised implementation, just the readability of the resulting source file.  Following the commit, lines 793-800 are just commented out, so will likely make it harder for future readers attempting to understand or maintain this class (why were these lines commented out? why were they there in the first place?)  So I am suggesting either removing those lines altogether (my preferred option) or adding some comment text nearby that explains why they're no longer required.

attipaci: Oh yes, sorry -- I misunderstood. Yes, I meant to remove the commented lines, just forgot... Thanks! :-D

attipaci:
Note, that the decrease in coverage is due to elimination of code, and not decrease of test cases. It's just that the same uncovered parts now constitute a larger fraction of the code in the affected files. Nevertheless, I'll see if I can compensate by additional test cases covering new things also...

Issue #189, opened 2021-10-29
Improve hierarchy and implementation of BufferedFile and related classes.


attipaci:
`BufferedFile` is currently implemented in a messy way. Not only it is hard to follow what it really does, but also, it's badly entagled with the `BufferEncoder` and `BufferDecoder` classes, whereby reads and writes are referenced both ways from both classes. Also, `BufferedDataInputStream` and `BufferedDataOutputStream` implement their own buffering when Java provides the buffering capability readily for these types. The same encoder and decoder classes should be used universally to read/write FITS to/from streams and files...

The files are also very poorly named, since all of them really support reading/writing FITS data format only. Specifically, characters should be treated as 1-byte ASCII, and booleans should be `T`, `F`, or 0 for `null`. So, they should be named `FitsFile`, `FitsEncoder`, `FitsDataInputStream` etc. rather then generic Buffer/Buffered nomenclature, which can be misleading. We cannot abandon the old names, but we can introduced better names, and deprecate the old ones.

attipaci:
The reorganized IO hierarchy and reimplementation of the file buffering resulted in a significant improvement in performance, typically by a factor of 2-4. Below are the detailed benchmarks from the test suite before and after the new IO classes:

__Before__ (old IO classes):

```
Standard I/O library: java.io.RandomAccessFile
  RAF Int write: 296.2962962962963
  RAF Int read:  727.2727272727273
  SyncRAF Int write: 320.0
  SyncRAF Int read:  800.0
Standard I/O library: java.io.DataXXputStream
                      layered atop a BufferedXXputStream
  DIS Int write: 48929.66360856269
  DIS Int read:  94955.48961424333
  DIS Int write: 117216.11721611721
  DIS Int read:  93294.46064139942
New libraries:  nom.tam.BufferedDataXXputStream
                Using non-tiledImageOperation I/O
  BDS Int write: 81218.27411167513
  BDS Int read:  66945.60669456067
New libraries: nom.tam.util.BufferedFile
               Using tiledImageOperation I/O methods.
  BF  Dbl write: 214046.8227424749
  BF  Flt write: 130612.24489795919
  BF  Int write: 140969.1629955947
  BF  Lng write: 219178.08219178082
  BF  Sht write: 80402.01005025125
  BF  Chr write: 44321.3296398892
  BF  Byt write: 1142857.142857143
  BF  Boo write: 34188.03418803419
  BF  Dbl read:  193353.4743202417
[...]

PR #192, opened 2021-11-02
Cleaned up the messy FITS IO hierarchy, and improved overall IO performance

attipaci:
Cleaned up the FITS IO hierarchies:

- Classes that are specific to the FITS binary format now have names starting with 'Fits', such as `FitsFile`, `FitsInputStream`, `FitsEncoder` etc. The old names of `BufferedFile`, `BufferedDataInputStream`, `BufferEncoder` etc. are retained for compatibility as deprecated subclasses of the newly introduced ones that replace them.
- The FITS-specific IO classes build on generic high-throughput IO base classes, such as `ArrayDataFile`, `ArrayInputStream`, `ArrayEncoder` -- from which one may derive other high-throughput IO classes that are not FITS specific.
- Both streams and files use the same encoder/decoder objects to translate between Java objects and binary representations. The encoding/decoding happens in a separate NIO buffer, which is one of the reasons for the apparent performance gain.
- The new IO classes are typically 2-4 times faster than the old implementations.
- The implementations of common read/write functions (such as `read(byte[], int, int)` have been changed to match the contract of `InputStream` and `RandomAccessFile` more closely and consistently. For example `read(byte[], int, int)` returns -1, rather than throwing an `EOFException` if there is nothing left to read from the input. The more consistent behavior means less surprises for devs who typically expect these methods to work the same way as for the built-in Java classes.
- Revised when and how exceptions are thrown in IO operations. The changes typically affect the internal use of IO classes and should not really affect the end user, who is typically unexposed to a lot of the affected IO API anyway.
- The `readLArray()` contract of `ArrayDataInput` explicitly required that the read should throw an `EOFException` if the supplied object could not be fully populated. However, that's now what was implemented. To reconcile, I changed the contract to codify the existing behavior instead of the idealized behavior, and added a new `readArrayFully()` method which does in fact throw an `EOFException` in incomplete reads. This new function is now used internally. 
- `FitsHeap` has been changed to be truly random accessible, rather than the one-directional read before, which required creating new streams every time one needed to reference backwards on the heap.
- Various smaller improvements to handling IO-related things, but careful to keep the old API intact, except where the old API was itself inconsistent or made no sense.


attipaci:
I'll work on bringing coverage back up before merging...

attipaci:
@mbtaylor, see the 3rd bullet point in the PR summary, and also Issue #189, which has more detail on the benchmarks. But as a quick summary here:

- I used the benchmarks that were part of the test suite (with slight modifications, such as running the tests longer to get better timing on them). The quoted values are quoted directly from the updated unit test output.
- Quoted from the PR summary: "Both streams and files use the same encoder/decoder objects to translate between Java objects and binary representations. The encoding/decoding happens in a separate NIO buffer, which is one of the reasons for the apparent performance gain."
- Some of the performance gain might also come from having streamlined the read/write methods, with more efficient loops and using more direct calls.

PR #202, opened 2021-11-18
1.16.0-rc3

attipaci:
- Move and hide `checkTruncated()` method to leave `ArrayDataInput` interface unperturbed. (Thanks to @mbtaylor for finding it as a potentially breaking change that was not necessary)

### DIFF
--- a/src/main/java/nom/tam/fits/Fits.java
+++ b/src/main/java/nom/tam/fits/Fits.java
@@ -1016,12 +1016,12 @@ public class Fits implements Closeable {
      * @since 1.16
      */
     static boolean checkTruncated(ArrayDataInput in) throws IOException {
-        if (!(in instanceof ArrayDataFile)) {
+        if (!(in instanceof RandomAccess)) {
             // We cannot skip more than is available in an input stream.
             return false;
         }
         
-        ArrayDataFile f = (ArrayDataFile) in;
+        RandomAccess f = (RandomAccess) in;
         long pos = f.getFilePointer();
         long len = f.length();
         if (pos > len) {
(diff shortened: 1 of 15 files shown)


mbtaylor:
Turns out it's not just `checkTruncated`; there are quite a few methods that have appeared on the `ArrayDataInput` interface, also `ArrayDataOutput`.  Fiddling with the source to try to get my existing code to build against this branch I've so far encountered:
```
    int read(Boolean[] buf)
    int read(Boolean[] buf, int offset, int size)
    void readArrayFully(Object o)
    int read()     // from InputReader
```
and in ArrayDataOutput I see
```
    void write(Boolean[] buf)
```
then I stopped counting; I don't know if other interfaces are affected too.

mbtaylor:
Possibly I could work round the interface additions, I haven't assessed how much effort that would involve.

attipaci:
@mbtaylor, I've added default methods for most of them. In some cases the default method is FITS-specific (like in case of `readArrayFully()`, which calculates the object's byte size assuming FITS encoding -- e.g. for `String`s). Others, like `writeBoolean(Boolean)` take a different default approach, e.g. for writing `null` as -1 by default. But then again, it does not matter since anyone implementing these interfaces outside of the library would not have used these new methods before, and if they did define methods by the same name/signature then those would anyway override the default. So, in essense there is little consequence of what these default methods are as long as they exist.

The one new interface method I am reluctant to add a default implementation for is `ReadWriteAccess.length()` (inherited by `RandomAccess`). If someone did implement their own nom-tam-fits `RandomAccess`, they probably already defined a `length()` method, and if not, they will just have to... I think that in that particular case defining a default implementation will likely to lead to trouble down the line somewhere...

mbtaylor:
Thanks for the changes. Indeed, I did already have a `length()` implementation. The other new method for which I don't see a default implementation is `read()`, inherited from `nom.tam.util.InputReader`.  I can provide implementations for that if you don't want to have a default implementation.

attipaci:
Oh, yes, that's another new one -- and `OutputWriter.write()` -- which are meant to provide a common interface for stream-like and file-like access (it's sort of surprising that Java does not have common interfaces for these built in...). A default implementation is not something that would work for these, so if you can implement these on your side that'd be great!

PR #282, opened 2022-04-14
Incremental HDU writing with automatic primary or extension HDU setting.

attipaci:
`BasicHDU.write()` now checks if the HDU is to be written to the start of the output (via the `FitsOutput` interface, implemented by both `FitsFile` and `FitsOutputStream`), and accordingly calls `setPrimaryHDU()` to create nearly fool-proof Fits files even if HDUs are written incrementally. 

However, `BasicHDU.write()` will allow writing a HDU to the head of stream, even if it cannot be made primary (such as a `BinaryTableHDU`), in order to better preserve compatibility with earlier behavior, and to allow writing incomplete Fits 'snipplets' if desired.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -8,7 +8,7 @@
 	</parent>
 	<groupId>gov.nasa.gsfc.heasarc</groupId>
 	<artifactId>nom-tam-fits</artifactId>
-	<version>1.16.2-SNAPSHOT</version>
+	<version>1.17.0-SNAPSHOT</version>
 	<packaging>jar</packaging>
 	<name>nom.tam FITS library</name>
 	<description>Java library for reading and writing FITS files. FITS, the Flexible Image Transport System, is the format commonly used in the archiving and transport of astronomical data.</description>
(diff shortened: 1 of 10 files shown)


attipaci:
Will be adding test cases next to fill coverage holes...

keastrid (changes requested):
Sorry for the delay.

Gave this a quick test, and its not fully working as-is: it seems `OutputEncoder` isn't updating `count` everywhere, such as in the various `put` methods. A simple fix that worked in testing was to add `count += bytes;` to `need`, though maybe doing it in `flush` would be better?

image

attipaci:
Thanks for checking!

I think the better solution is to change `OutputEncoder.getCount()` to:

```java
public long getCount() {
    return count + buf.buffer.position();
}
``

This way it always returns the actual number of bytes that have either been flushed already or are waiting in the holding buffer.

keastrid:
That misses the code path when flushing, which resets the buffer position to 0. In `flush`, adding `count += n;` fixes the issue.

attipaci:
I did miss that, thanks!

PR #287, opened 2022-04-24
Dynamically adjust the number of Forked VM based on the number of CPU cores

CAIWan1998:
Maven will run all tests in a single forked VM by default. This can be problematic if there are a lot of tests or some very memory-hungry ones. We can fork more test VM by setting `<fork>1.5C</fork>`.

=====================
If there are any inappropriate modifications in this PR, please give me a reply and I will change them.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -430,7 +430,7 @@
 				<artifactId>maven-surefire-plugin</artifactId>
 				<version>3.0.0-M6</version>
 				<configuration>
-					<forkCount>1</forkCount>
+					<forkCount>1.5C</forkCount>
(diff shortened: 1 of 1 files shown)


attipaci:
Thanks! That makes sense. And, it looks like the tests passed just fine. However, before I merge I want to think a bit more about it...

There are a number of tests which use the temporary same file name (for writing and reading back). Blind parallelization of these tests could result in ugly race conditions. We can avoid them by explicitly making sure each test uses a unique file name for temporary test files, but it will take some careful checking of the test code to identify and change all such cases...

In any case thanks again for suggesting a worthy change, and welcome to the nom-tam-fits community!

PR #291, opened 2022-04-26
Add support for AstroImageJ's `ANNOTATE` keyword

keastrid:
Currently, nom.tam has issues with the `ANNOTATE` keyword as it does not exactly follower FITS spec. Multiple are allowed, and they contain a string value.

This add support for correctly reading and writing the `ANNOTATE` keyword.

Not entirely sure if I implemented `IFitsReader` correctly, but the changes suffice for something to work.


attipaci:
Hi @keastrid!

Thanks for the PR. I did take a look and I have a few suggestions before I can merge (trivial stuff, really):

1. `nom.tam.fits.test.EnumHeaderTest` fails on line 224. You simply need to bump the expected value up by one to 15, now that you added an extra source.
2. Please refactor `nom.tam.fits.header.Aij` to `nom.tam.fits.header.extra.AIJExt` for consistency. (All keyword extensions go into the `nom.tam.fits.header.extra`, and the naming convention is to keep acronyms capitalized, followed by 'Ext').
3. You should add some test cases, e.g. in `HeaderTest`. At minimum you want to check (a) that you can add multiple instances of this keyword without encountering an exception, and (b) that a card created with the `ANNOTATE` IFitsHeader keyword is indeed a comment-style card (`HeaderCard.isCommentStyleCard()`). You can add any number of further checks that you deem appropriate. Preferably, create multiple small tests, each with just 1 or 2 assertions, rather than one big test with dozens of assertions.

That's all that comes to mind at first peek. I'll take a look again when you are ready with an update for this PR. :-)

And, thanks again!

keastrid:
I've made the requested changes.

The test for `isCommentStyleCard` isn't working. I've tried passing `NONE` and `null` in the constructor for the card, but neither seem to do the trick, so for now I've disabled that test.

attipaci:
I'm not too surprised. In principle, one should always create comment cards using `HeaderCard.createCommentStyleCard(String, String)`. It is also wrapped by `Header.insertCommentStyle(String, String)` and `Header.insertCommentStyleMultiline(String, String). HeaderCards created by the constructor are always key,value cards, even if the value is `null` (which the Fits standard allows for)...

I think, ultimately we can use valueType=null to designate comment-style IFitsHeaders (e.g. as `COMMENT` already does), and make header check and use that to act approriately. However, Header.addValue(IFitsHeader...) clearly does not make sense for comment cards with no assigned value (only comment). So, in the end you should be using one of the `CommentStyle` calls mentioned above to create `ANNOTATE` cards properly.

I bet that if you update your tests to create cards that way, you'll be fine. I think after I merge your PR, I'll update `Header.addValue(IFitsHeader...) to throw an exception if you try to use it with cards that have valueType=null... (and similarly, at other places where IFitsHeaders are used...). But that's for later...

Let me know what you think...

attipaci:
BTW, your PR does not have complete diff coverage. You can check the Codecov report above (click the link) to see what lines need to be covered. As a general principle, merging PRs requires 100% diff coverage, and no decrease of overall coverage... I know it's annoying, but it's a good thing :-)

attipaci:
PS, I changed the repo config s.t. the CI should run automatically for you, so you get coverage reports after every commit to help you fill in those holes...

keastrid:
I'll take a look at the diff coverage.

My confusion with the comment-style cards comes from the creation of the `ANNOTATION` card in `AIJExt` and what happens when that card is read from a fits file.

attipaci:
Reading should be fine. If there is no '=' in character 9 (counting from 1), then the parser marks the cards as a comment-style card (type=null). If that does not happen, we have a problem somewhere...

keastrid:
Woops - got confused by the duplicates being allowed like with `COMMENT`. It is NOT a comment style card, sorry for the confusion!

attipaci:
Well, in that case, `ANNOTATE` fundamentally violates the FITS standard. The current FITS specification document states (on p.7):

> The mandatory FITS keywords defined in this Standard must not appear more than once within a header. All other keywords that have a value should not appear more than once.

I think nom-tam-fits should warn about duplicates for this card, if it is not used as a comment-style card. (You can disable the logging for such violations, of course).

I think AstroImageJ should revisit their use of ANNOTATE to define it as a comment-style keyword (no `=` in byte 9). They can continue to support their old convention with the `=` there, but I don't see why anyone else should support something that is blatantly in violation of the FITS standard...

It's not my final word, but I will need a lot more convincing to see why we should exempt a poorly conceived `ANNOTATE` keyword from the standard rules...

keastrid:
Yes, it breaks the ideal in the FITS standard, but I would argue that only the standard keywords actually enforce that ("must" vs "should"); the standard keywords also define the "form" of the FITS file, so allowing duplicates would be meaningless or necessitate a more complex resolution strategy whereas 3rd party keywords don't need that - having multiple `FOO`s is harmless to the file itself, and only has meaning to the readers aware of the keyword.

Even if it was changed to a comment style card, there are years of older images that exist with the old one.

I believe in the original testing (and currently) nom.tam warns about the duplicate card in the log, but then eats the duplicate cards and only returns one of them with the header. As far as I can tell, there isn't a way to allow that on AIJ's side without using a separate reader.

This is in contrast to other readers that will return the full out-of-spec header.

attipaci:
I don't disagree that we should allow multiple keywords -- and we already do. But, my understanding of _should_ is that yes you can deviate from it, but a warning about it is absolutely justified -- because you really _should_ not have done that in the first place. The `Logger` warnings don't not break anything, and if it's annoying from a aesthetic point of view, then you can mute those `Logger` warnings. If aesthetics is really such an issue then AstroImageJ should fix the messed up headers they produce. That to me is a bigger aesthetic issue...

Nevetheless, I still think we should add your `AIJExt` because it is a keyword after all that is commonly used. But, I don't think we should give it a completely free ride in the header parsing for duplicates. My preferred way of dealing with it is either

 -  having a static switch in `Header` that could be used to suppress only the duplicate warnings alone, and nothing else (e.g. `void Header.suppressDuplicateWarnings(boolean)` and `boolean Header.isDuplicateWarningsEnabled()`), and/or
 -  always suppressing multiple `Logger` messages about the same duplicate keywords, so you would get just one `Logger` warning about `ANNOTATE` dupes and not a thousand...

I don't think there is a truly compelling argument to give special treatment to `ANNOTATE`. My worry is that if we go down that road, we'll have to start accommodating everyone's quirky FITS deviants, and we'll end up making a mess out of it with countless `if`, `else if`, `else` statements -- and noone can predict how the library will behave with certainty in any particular case. I think for the sake of staying consistent, I vote for no special treatments over giving everyone their own special treatment...

Let me know how you feel about it.

attipaci:
As for the behavior of 'eating' the duplicates, the FITS standard also says (the sentence immediately after the quote further above:

>  If a keyword does appear multiple times with different values, then the value is indeterminate.

So, yes, if `ANNOTATE` appears multiple times, it's value really should be treated indeterminate, by definition. However, nom.tam.fits does actually keep track of the alternate versions of the duplicate cards. While `Header` will report only the first occurrence in the regular way, you can call `Header.getDuplicates()` to retrieve all occurrences as a list (so even order is maintained!).

Hopefully that answers your concerns about processing those multiple `ANNOTATE` keywords without much extra effort... 

Does that make sense?

keastrid:
Hmm, I see what you mean with the undetermined value from the spec.

For implementation of breaking keywords, it doesn't need to be a massive `if-else` chain - a simple map/list lookup should suffice, and if made accessible the API caller could then add their various keywords to the map to have the duplicates passed through. The FITS standard has been around long enough I doubt AIJ is the only one that breaks it.

Something like:
```java
public class SpecialCardHandling {
	//not necessarily having these here, but unifies the impl.
	private static List<String> DUPLICATE_ALLOWED = new List({"CONTINUE", "HISTORY"});
	
	public static void allowDuplicateForKey(String key) {
		//add to list
	}
	
	//reset method to put it back to continue, history
	
	//remove key from list method
	
	public static boolean isDuplicateAllowed(String key) {
		//list lookup
	}
}

//change if-statements of duplicate checks to SpecialCardHandling.isDuplicateAllowed()
```

My concern isn't really with the logging.

Is it intentional for `Header.dumpHeader` to not include the duplicates? 

From what I can see these are the implementation issues for with handling duplicate keywords:
- Converting the header w/ duplicates to a newline delineated string ala `dumpHeader` (simple enough to work around)
- Adding duplicate cards to a header for writing (currently replaces the previous occurrence's value)

My understanding is that as it stands reading a file then writing it out would also remove the duplicate keywords (I don't see the duplicates list being used anywhere) - I would expect such an operation to be lossless, or at least opt-in for lossy.

attipaci:
Right, the philosophy of the library is that we can be tolerant to wacky things when reading FITS files, but we should only produce FITS files that conform to the standard (otherwise other tools will struggle with out quirks too). This is why re-writing an improper header (with duplicates) will make it proper (by removing the duplicates). That's absolutely what we expect the library to do. And, there is really no desire to change that...

However, there is a back door, which is to use cursor access for editing headers. While `Header.addValue()` will replace existing values, `Cursor.add(HeaderCard)` will allow you to add duplicates unchecked to your pleasure. So you could try to use that...

My problem with the special handling you suggested, is that it's fine for duplicates, but other quirks include having the = in the wrong place, or nor being followed by a space, or not liking string padding, or having lower case characters in keywords, etc. The possibilities for special handling are endless, and your class won't catch all. So, my vote is still agaist adding any special handling for your case...

I would consider exposing the `HeaderCardParser` class, so that you can also implement your own low-level, line-by-line reading of headers, rather than just going with `Header.read()`... That might help your use case as well. Then, you can write your own library, on top of nom-tam-fits, to handle `ANNOTATE` to your liking, without asking nom-tam-fits to deal with a special case...

Would that be reasonable?

attipaci:
PS. If you really just need to dump the contents of these funky headers, you only really need the library to give you the file offset (`BasicHDU.getFileOffset()`), then you repeatedly do a standard `read()` into a `byte[80]` buffer, until the the first 8 bytes are 
`END     `. Then you can print it to an output using `println(new String(buf))`, or do whatever else you want. You don't need the library to interpret the header for you at all. But, if you want the header in a digested form (as per the FITS specification), the library gives you the tools, as far as your header conforms to the specs (and more)...

attipaci:
To help your cause I'm adding a

```java
public void validateFor(BasicHDU<?> hdu, FitsOutput out) throws FitsException 
```
call to `Header` (on my duplicate-warnings branch), which does the same header checking including setting primary as the standards `Header.write()` called by `BasicHDU.write()` would do. So, after calling it, you can simply write the cards out. You could use it, e.g. with an empty header to set the required keywords. Then you can write out the cards from it before `END`. After which you can write whatever cards you like in whatever order. Does that sound like a reasonable solution to you?

attipaci:
I'll close this issue for lack of activity. We can always reopen it again if there is a reason for it.

Issue #341, opened 2022-10-07
Rare NPE with writing of FPACKed files.


keastrid:
Sometimes when writing an FPACKed HDU, an NPE is thrown due to compressOptions being null, despite being set.

This isn't something I can reliably reproduce (only occurred once for me), though another was able to reproduce it several times with the export of multiHDU files.

The solution we ended up with that seems to work is changing `nom.tam.image.compression.tile.TiledImageCompressionOperation#initializeCompressionControl` to 

```java
private void initializeCompressionControl() {
        if (this.compressorControl == null || imageOptions == null) {
            this.compressorControl = CompressorProvider.findCompressorControl(this.quantAlgorithm, this.compressAlgorithm, getBaseType().primitiveClass());
            if (this.compressorControl == null) {
                throw new IllegalStateException("Found no compressor control for compression algorithm:" + this.compressAlgorithm + //
                        " (quantize algorithm = " + this.quantAlgorithm + ", base type = " + getBaseType().primitiveClass() + ")");
            }
            initImageOptions();
        }
    }
```

[Log](https://gist.github.com/keastrid/a23e3b6ba5719f78fe71208c40803d40)

attipaci:
Thanks for the solution @keastrid. I'll make sure to include that in the next update (around December 15). Do you want to make a PR with the fix, or shall I?

keastrid:
Went ahead and made a PR.

PR #342, opened 2022-10-08
Fix GH-341

keastrid:


attipaci (commented):
I wonder if a slightly different change would be better:

```java
private void initializeCompressionControl() {
        if (this.compressorControl == null) {
            this.compressorControl = CompressorProvider.findCompressorControl(this.quantAlgorithm, this.compressAlgorithm, getBaseType().primitiveClass());
            if (this.compressorControl == null) {
                throw new IllegalStateException("Found no compressor control for compression algorithm:" + this.compressAlgorithm + //
                        " (quantize algorithm = " + this.quantAlgorithm + ", base type = " + getBaseType().primitiveClass() + ")");
            }
        }
        if (imageOptions == null) {
            initImageOptions();
        }
    }
```

This way, `compressorControl` is not reinitialized if it exists, and only the `imageOptions` is null...

What do you think? Also, would you be willing to add a test case for initializing with imgaeOptions being both null / non-null to check for both scenarios in the future (provided such a test can be constructed given visibilities/permissions).

keastrid:
I've made the requested change, it should work fine.

I don't have time to make a test in the immediate future (our own testing setup still needs work), but I don't think it is possible in this case, as there seems to be no way to access imageOptions or force it to be null.

attipaci:
Thanks for making the change. Yes, I did suspect that testing for this might be difficult for this change specifically. So, let's not worry about it. :-)

attipaci (approved):

PR #350, opened 2022-11-23
Issue 349

attipaci:
It seems like GZIPed image decompression is not taking ZZERO and ZSCALE values into accout...


attipaci:
I'm going to need to fill the coverage holes before we can merge. In the meantime @bogdani and @keastrid can check out the functionality...

PR #352, opened 2022-12-15
Changes to allow alternate random access into Fits files.

### LINKED ISSUE #351: Allow for alternate RandomAccess implementations
To facilitate random access over the network (byte range requests), it would be nice if the `Fits` class could support alternate sources of random access data.

at88mph:
Fixes #351 

Some typo and deprecation fixes as well.


src/main/java/nom/tam/util/RandomAccessDataObject.java
```diff
@@ -0,0 +1,84 @@
...
+/**
+ * Minimal interface for underlying data object that supports random access. The
+ * methods defined here are those used by FitsFile to access a RandomAccessFile.
+ * The RandomAccessFileExt class adds this interface to RandomAccessFile, but other
+ * systems could provide an alternate implementation of this interface to access
+ * an arbitrary FITS data object.
+ * 
+ * @author pdowler
+ */
+public interface RandomAccessDataObject extends Closeable {
```
attipaci, line 48: I'd suggest to name this differently. Since it's an interface, I'd not put `Object` in the name. How about `RandomAccessible` instead? (It's a pity that we also have a `nom.tam.util.RandomAccess` interface, which in restrospect would have been better called something more specific, like `RandomFitsAccess`, but oh well...)

src/main/java/nom/tam/util/RandomAccessDataObject.java
```diff
@@ -0,0 +1,84 @@
...
+
+    // write methods from RandomAccessFile
+    
+    void setLength(long l) throws IOException;
+
+    void write(byte[] bytes, int i, int i1) throws IOException;
+
+    void write(byte[] bytes) throws IOException;
+
+    void write(int i) throws IOException;
```
attipaci, line 81: Declared in `OutputWriter` via `ReadWriteAccess`

src/main/java/nom/tam/fits/Fits.java
```diff
@@ -1266,9 +1297,9 @@ public static long checksum(final byte[] data) {
      *              the end-of-file, otherwise <code>false</code>.
      * @throws IOException  if there was an IO error accessing the file or stream.
      * 
-     * @see #skip(long)
-     * @see #skipBytes(int)
-     * @see #skipAllBytes(long)
+     * @see ArrayDataInput#skip(long)
+     * @see ArrayDataInput#skipBytes(int)
+     * @see ArrayDataInput#skipAllBytes(long)
```
attipaci, line 1302: Thanks! :-D

src/main/java/nom/tam/util/RandomAccessDataFile.java
```diff
@@ -0,0 +1,49 @@
...
+import java.io.FileNotFoundException;
+import java.io.RandomAccessFile;
+
+/**
+ * Simple extension so randomAccessFile implements internal
+ * RandomAccessDataObject.
+ * 
+ * @author pdowler
+ */
+public class RandomAccessDataFile extends RandomAccessFile implements RandomAccessDataObject {
```
attipaci, line 44: Do we really need a new class here? Could we not simply make the existing `RandomAccessFile` implement the new interface (which it already does)? I'd prefer to keep it simpler...

attipaci (commented):
Hi!

Thanks for this PR. It makes sense and I like the idea. However, I'd suggest to simplify it a little and integrate it better with the existing classes/interfaces. See my specific comment below. I look forward to your response and the next iteration...

src/main/java/nom/tam/util/RandomAccessDataObject.java
at88mph: How about `RandomDataAccess`?

src/main/java/nom/tam/util/RandomAccessDataFile.java
at88mph: It's not necessary.  It just provided a nice sample implementation.

src/main/java/nom/tam/util/RandomAccessDataObject.java
attipaci: Well, my issue with that is that there is nothing 'Data' specific in the interface either. It's simply the methods of `RandomAccessFile` extracted into an interface. It is very 'File' specific though since it contains a `getFD()` method. So, maybe `RandomAccessFileIO`? (That would capture that it is both a  random-access interface and file-specific...)

at88mph:
Rework submitted.

attipaci:
Hi @at88mph,

I will check and merged this into 1.18.0 (scheduled for 6/15) after 1.17.1 is released (sometime soon, I hope)

attipaci:
Hi @at88mph,

Thanks for making the changes. It's looking better now. I expect to start merging 1.18.0 PRs around April. In the meantime, can I ask you to add unit tests as needed to keep overall coverage undiminished, and ideally also to provide 100% coverage for your additions? Thanks in advance.

at88mph:
Ah, I thought I took care of that.  Sure, I will get this fixed up.

attipaci:
@at88mph, I resolved the conflicts with the mainline above. You will have to pull it. I don't think I deleted or modified any code while resolving the conflict, but I did make changes to some of the Javadoc, mainly in `Fits.java` to make the new constructor more consistently documented with the others, and to cross-reference your additions.  Please check to make sure all is as you expect. :-)

The coverage decrease is pretty small (-0.04%), so it just needs a few more lines to be covered with tests. Almost there...

at88mph:
OK, I think I've managed to include tests for added logic.  Thanks!

attipaci (approved):

PR #356, opened 2023-01-10
Stream feature

### LINKED ISSUE #355: Enable streaming output for tiling
Related to #282 and #100 

Slicing a tile from a large FITS file will produce an `OutOfMemoryException`, which is particularly problematic when serving data on the web.  Ideally, a cutout `Fits` object could be constructed from a source `Fits` object without reading anything but the base HDU object(s) into memory.  A pseudo code example of what I'm proposing:

```java
     Fits source = new Fits(myFile);
     Fits output = new Fits();
     ImageHDU imageHDU = source.getHDU(1);
     Header tileHeader = adjustHeaderToTile(imageHDU.getHeader()); // Correct the CRPIXn and other headers to match the new dimensions.
     int[] tileStarts = new int[]{10, 10};
     int[] tileLengths = new int[]{45, 60};
     if (overlap(imageHDU.getData())) {
         StreamingTileImageData streamingTileImageData = new StreamingTileImageData(tileHeader, imageHDU.getTiler(),
                                                                                    tileStarts, tileLengths);
         output.addHDU(FitsFactory.hduFactory(tileHeader, streamingTileImageData));
     }
     output.write(outputStream);  // The cutout happens at write time!
```

at88mph:
Fixes #355 

Add ability to stream data out for cutouts.

 - Added `StreamingTileImageData` to support writing tiles only when `write()` is called.
 - Added support for `ArrayDataOutput` where an `Object` was used to write out a primitive array.


attipaci:
Hi @at88mph,

Sorry for the slow response. I'm flooded with other stuff these days. In any case, this being a new feature, it will go into the next version (1.18.0, planned for a 6/15 release). Before that, there is a blocking bug for 1.17.1 in the way tile compression is implemented that needs to be fixed before I'll start merging PRs for 1.18.0. I hope I can count on your continued patience... :-).

attipaci:
@at88mph, Nicely done! 

+1

pom.xml
```diff
@@ -215,7 +215,8 @@
 				<artifactId>maven-compiler-plugin</artifactId>
 				<version>3.11.0</version>
 				<configuration>
-					<release>8</release>
+					<source>1.8</source>
```
attipaci, line 218: I am pretty sure we want to revert this one to use `<release>` instead of `<source>` and `<target>`. If I recall correctly the newer Java compilers (Java 9+)  no longer support separate `source` and `target` flags, and this will break build on all of the more recent compilers. The new compilers use `release`.

Of course, using `release` will break compilation with Java 8. But, if we must break something (and we don't have a choice here), I'd rather not break the more modern versions. Rather, you probably want to start compiling using `javac` from Java 9 or newer...

attipaci (changes requested):
Overall it looks good, and since it's new code I don't have strong opinions about choices that it makes. :-D

There is a critical change that you made in `pom.xml` that will need to be reverted I think. This has caused problems before. Other than that it's good to go.

pom.xml
attipaci: Here is an [article](https://saker.build/blog/javac_source_target_parameters/index.html) that describes the issue better than I can...

at88mph: Are you using JDK 9 now?  I normally use JDK 11 but there are conflicts with the current dependency tree.  I will revert this change.

at88mph: Done.

attipaci: I'm using 11 for the CI (and 17 for myself). 9 is simply the lowest version that works...

at88mph: Excellent.  So we can say that `nom-tam-fits` supports JDK 17?  I'd rather keep using that if so.  Thanks!

attipaci (approved):

PR #380, opened 2023-03-16
Use ZDITHER0 and seed dither random generator for each tile as prescribed

attipaci:
This PR addresses two related issues:

1. `ZDITHER0` was not used (written or read) to record or seed the random generator when dithering.
2. Initializing the dither random generator an a per-tile basis using `ZDITHER0` as prescribed by the FITS standard.


attipaci:
@keastrid,

I m hopeful that this resolves your issue. The test image you provided now decompresses consistent with what ds9 shows for the compressed image. Please check it out and let me know if you are happy with the way things are.

keastrid:
This seems to work, thanks!

PR #381, opened 2023-03-17
Try alternative coveralls upload (tokenless)

attipaci:
The current upload requires a repo secret, so it does not upload coverage for PRs. There is a pre-release version of a GitHub Actions that should allow tokenless coverage upload to Coveralls.io. Let's see if we can make that work...


attipaci:
The coveralls action still does not process jacoco.xml unfortunately (despite some threads claiming so). So, I'll put this PR to sleep for now. Maybe we'll revisit later when there is indication that things might change...

PR #388, opened 2023-03-30
Step feature

### LINKED ISSUE #387: Add ability to stride while tiling
Similar to `numpy`'s ability to stride (step) through an array in order to reduce the output size, the streaming tiler could step over a certain number of bytes while reading.

To visualize this, a typical array stride would look like:
`cutout = my_array[10:100:2]`

Where the length is 90 values long, while striding very second value.  The `cutout` array would be 45 values long.

at88mph:
Closes #387 

Adds tests and striding ability for the tilers.

### DIFF
--- a/src/main/java/nom/tam/image/ImageTiler.java
+++ b/src/main/java/nom/tam/image/ImageTiler.java
@@ -46,10 +46,13 @@ public interface ImageTiler {
 
     Object getTile(int[] corners, int[] lengths) throws IOException;
 
-    Object getTile(int[] corners, int[] lengths, int[] steps) throws IOException;
+    default Object getTile(int[] corners, int[] lengths, int[] steps) throws IOException {
+        throw new UnsupportedOperationException("Striding feature not yet implemented.");
+    }
 
     void getTile(Object array, int[] corners, int[] lengths) throws IOException;
 
-    void getTile(Object array, int[] corners, int[] lengths, int[] steps) throws IOException;
-
+    default void getTile(Object array, int[] corners, int[] lengths, int[] steps) throws IOException {
+        throw new UnsupportedOperationException("Striding feature not yet implemented.");
+    }
 }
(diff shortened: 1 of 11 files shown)


src/main/java/nom/tam/image/ImageTiler.java
```diff
@@ -46,6 +46,10 @@ public interface ImageTiler {
 
     Object getTile(int[] corners, int[] lengths) throws IOException;
 
+    Object getTile(int[] corners, int[] lengths, int[] steps) throws IOException;
+
     void getTile(Object array, int[] corners, int[] lengths) throws IOException;
 
+    void getTile(Object array, int[] corners, int[] lengths, int[] steps) throws IOException;
```
attipaci, line 53: Like above, add a default implementation throwing `UnsupportedOperationException`.

attipaci (changes requested):
Thanks for the nicely done PR. Just a few tweaks to maintain backward compatibility more closely... The changes are trivial, but they will require some tests also... Thanks in advance!

attipaci:
@at88mph, thanks for adding yet another nice feature! It's looking pretty good. I'll merge it once it's more back compatible with the previous API. :-)

src/main/java/nom/tam/image/StandardImageTiler.java
```diff
@@ -74,15 +71,15 @@ public static long getOffset(int[] dims, int[] pos) {
      * @param start   The starting corner values.
      * @param current The current offsets.
      * @param lengths The desired dimensions of the subset.
+     * @param steps   The desired number of steps to take until the next position.
      * @return <code>true</code> if the current array was changed
      */
-    protected static boolean incrementPosition(int[] start, int[] current, int[] lengths) {
-
+    protected static boolean incrementPosition(int[] start, int[] current, int[] lengths, int[] steps) {
```
attipaci, line 77: Ooops, I did miss this one before. It's another one where we should keep the old method signature as well as adding the new one.... The rest looks good to go :-)

attipaci (changes requested):
Just one more tiny back compatibility thing I missed before. Once that's done, I'll merge to master :-). Thanks for the great work!

attipaci (approved):

PR #391, opened 2023-04-01
Issue 390: Reinstate deleted compression classes

attipaci:
Reinstating Float/Double compression classes for Rice and HCompress, including dedicated option classes for these.


attipaci:
Ooops, this was branched from `header-only-hdu` instead of `master`. No biggy. We'll merge it after header-only-hdu...

PR #399, opened 2023-04-14
Fix CompressorProvider.NULL_PARAMETERS not playing nice when bundled.

attipaci:
Compression parameters have a doubly abstracted hierarchy, based on an interface, on top of which lies an abstract class providing another layer of abstract method stems that subclasses must all implement. In effect, the abstract class does what default methods should have been used for in the base interface. This double layering of abstract methods is unfortunate, but since this is how it has been since the compression classes were added in 1.15, we have no choice but to keep it that way, lest we want to risk back compatibility issues.

The problems start when bundling options (in PR #350) that implement only the base interface, but not the additional abstract methods of the abstract class atop the interface, such as was the case for `CompressorProvider.NULL_PARAMETERS`. The problem is easily fixed by deriving these internal parameters from the abstract class rather than the base interface.

We also need to warn users to avoid implementing the base interface directly externally to the library, and instead implement the methods of the abstract `CompressParameters` class instead.

### DIFF
--- a/.github/workflows/maven.yml
+++ b/.github/workflows/maven.yml
@@ -58,7 +58,6 @@ jobs:
       with:
 #        token: ${{ secrets.CODECOV_TOKEN }}
         working-directory: ./nom-tam-fits
-        files: ./target/site/jacoco/jacoco.xml
         fail_ci_if_error: false
         flags: unittests
         name: codecov
--- a/.github/workflows/maven.yml
+++ b/.github/workflows/maven.yml
@@ -57,7 +57,6 @@ jobs:
       continue-on-error: true
       with:
 #        token: ${{ secrets.CODECOV_TOKEN }}
-        working-directory: ./nom-tam-fits
         fail_ci_if_error: false
         flags: unittests
         name: codecov
--- a/codecov.yml
+++ b/codecov.yml
@@ -7,6 +7,10 @@ coverage:
   range: "80...100"
   status:
     patch: off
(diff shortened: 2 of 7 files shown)


keastrid:
Thanks for the quick fix!

This seems to fix the CCE, however pixel data is being returned as all 0s for the test image in the original issue.

I have yet to find the cause of this, I don't believe it is how we are using nom.tam as other fpacked files still work, and prior to 1.18 the test image functioned.

attipaci:
Oh well, one baby step at the time. I'll see what I can make of the all zeroes...

attipaci:
Hi @keastrid,

So, I think the problem stems from the header entry:

```
ZQUANTIZ= 'NONE    ' 
```

According to the FITS standard, `ZQUANTIZ` can only take values `No_DITHER`, `SUBTRACTIVE_DITHER_1` or `SUBTRACTIVE_DITHER_2`. If no quantization is used (as in your case, the ZQUANTIZ keyword should not be present. So, you can try remove that header keyword from your compressed FITS and see if that fixes the issue.

If so, I'll add some validation to the parsing of `ZQUANTIZ` so that it will treat non-standard values as if the keyword were not present, and perhaps log a warning about it...

Let me know if that does the trick...

keastrid:
Thanks, that did the trick!

I don't know why the dither keyword was present. Unfortunately, I don't recall how these files were compressed/produced.

PR #400, opened 2023-04-17
New ability to stream cutouts from compressed images

at88mph:
A bit of an ugly one.  This is to support cutting out of Compressed images to a Stream.  We have very large cubes compressed using `fcat` that need to support streamed cutouts.  This also includes some fixes for the re-introduced `Float` support.

### DIFF
--- a/src/main/java/nom/tam/fits/compression/algorithm/hcompress/HCompressorQuantizeOption.java
+++ b/src/main/java/nom/tam/fits/compression/algorithm/hcompress/HCompressorQuantizeOption.java
@@ -32,7 +32,6 @@ package nom.tam.fits.compression.algorithm.hcompress;
  */
 
 import nom.tam.fits.compression.algorithm.quant.QuantizeOption;
-import nom.tam.fits.compression.algorithm.rice.RiceCompressOption;
 
 public class HCompressorQuantizeOption extends QuantizeOption {
 
@@ -40,7 +39,7 @@ public class HCompressorQuantizeOption extends QuantizeOption {
         super(new HCompressorOption());
     }
 
-    public HCompressorQuantizeOption(RiceCompressOption compressOption) {
+    public HCompressorQuantizeOption(HCompressorOption compressOption) {
         super(compressOption);
     }
 
--- a/src/main/java/nom/tam/fits/compression/algorithm/hcompress/HCompressorQuantizeOption.java
+++ b/src/main/java/nom/tam/fits/compression/algorithm/hcompress/HCompressorQuantizeOption.java
@@ -32,11 +32,16 @@ package nom.tam.fits.compression.algorithm.hcompress;
  */
 
 import nom.tam.fits.compression.algorithm.quant.QuantizeOption;
+import nom.tam.fits.compression.algorithm.rice.RiceCompressOption;
 
 public class HCompressorQuantizeOption extends QuantizeOption {
 
     public HCompressorQuantizeOption() {
         super(new HCompressorOption());
+    }
+
+    public HCompressorQuantizeOption(RiceCompressOption compressOption) {
+        super(compressOption);
     }
 
     public HCompressorOption getHCompressorOption() {
--- a/src/main/java/nom/tam/fits/compression/algorithm/rice/RiceQuantizeCompressOption.java
+++ b/src/main/java/nom/tam/fits/compression/algorithm/rice/RiceQuantizeCompressOption.java
@@ -31,7 +31,6 @@ package nom.tam.fits.compression.algorithm.rice;
  * #L%
  */
 
-import nom.tam.fits.compression.algorithm.api.ICompressOption;
 import nom.tam.fits.compression.algorithm.quant.QuantizeOption;
 
 public class RiceQuantizeCompressOption extends QuantizeOption {
@@ -40,7 +39,7 @@ public class RiceQuantizeCompressOption extends QuantizeOption {
         super(new RiceCompressOption());
     }
 
-    public RiceQuantizeCompressOption(ICompressOption compressOption) {
+    public RiceQuantizeCompressOption(RiceCompressOption compressOption) {
         super(compressOption);
     }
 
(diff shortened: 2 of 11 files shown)


src/main/java/nom/tam/image/compression/hdu/CompressedImageHDU.java
```diff
@@ -142,6 +144,48 @@ public CompressedImageHDU(Header hdr, CompressedImageData datum) {
...
+    }
+
+     /**
+     * Given this compressed HDU, get the original (decompressed) axes.
+     *
+     * @return the dimensions of the axis.
+     * @throws FitsException
+     *             if the axis are configured wrong.
+     */
+    public int[] getDecompressedAxes() throws FitsException {
```
attipaci, line 161: I wonder if this would be better named as `getImageAxes()`. To me that would make it a bit more intuitive as to what it represents...

src/main/java/nom/tam/image/compression/hdu/CompressedImageHDU.java
```diff
@@ -142,6 +144,48 @@ public CompressedImageHDU(Header hdr, CompressedImageData datum) {
...
+        return axes;
+    }
+
+    /**
+     * Obtain a header representative of a decompressed ImageHDU.
+     * @return Header with decompressed cards.
+     * @throws HeaderCardException
+     *          if the card could not be copied
+     */
+    public Header getGenericHeader() throws HeaderCardException {
```
attipaci, line 188: I wonder if this should have more limited visibility? Or do we want this exposed to the users? Also perhaps call it `getImageHeader()` or `getDescriptiveImageHeader()`?

src/main/java/nom/tam/fits/compression/algorithm/rice/RiceQuantizeCompressOption.java
```diff
@@ -39,6 +40,10 @@ public RiceQuantizeCompressOption() {
         super(new RiceCompressOption());
     }
 
+    public RiceQuantizeCompressOption(ICompressOption compressOption) {
```
attipaci, line 43: This may be fine, but if there is a way to restrict the argument to `RiceCompressOption` class it would be better, as it would prevent some misuse. Of cousre, if it's invoked from a generic code where the options class is not given, it may not help...

attipaci (changes requested):
Thanks for adding yet another nice feature. I'm not going to be very picky here, but I do have  few suggestions that you may take or leave. Either way, I'll be ready to merge this once you are happy with it yourself.

src/site/markdown/index.md
```diff
@@ -340,8 +386,53 @@ If the pixels are ints, that’s  an 8 GB file.  We can do
...
 The `reset()` method causes the internal stream to seek to the beginning of the data area.
-If that’s not possible it returns false. 
+If that’s not possible it returns false.
+
+#### Random Access
+It is possible to specify alternate random access by implementing the `RandomAccessFileIO` interface.  This is useful
+when a Fits file isn't necessarily on disk, but rather in object storage on the web, for example, or any other
+network access.
 
+```java
```
attipaci, line 396: I'm not sure we need to spell out in the README how to implement the interface, given especially that the example implementation is pretty much trivial (yet long).

src/site/markdown/index.md
```diff
@@ -340,8 +386,53 @@ If the pixels are ints, that’s  an 8 GB file.  We can do
...
+import nom.tam.util.RandomAccessFileIO;
+
+import java.io.FileDescriptor;
+import java.io.IOException;
+import java.nio.channels.FileChannel;
+
+public final class S3RandomAccessFileIO implements RandomAccessFileIO {
```
attipaci, line 403: Perhaps just keep this line in your example, and replace the body with just ellipses (`...`). Also, currently you reference `S3RandomAccessFileIO` in the README earlier, so it's too late to specify it here what it is. Either define it earlier (preferably just before the examples that refer to it), or use a more vanilla RandomAccess implementation in those examples....

attipaci (changes requested):
Some additional suggestion for the README....

src/main/java/nom/tam/fits/compression/algorithm/rice/RiceQuantizeCompressOption.java
attipaci: Oh, and while at it, you probably want to do the equivalent for `HCompressQuantizeOption` also (regardless of which method signature you settle on.)

src/main/java/nom/tam/image/compression/hdu/CompressedImageHDU.java
at88mph: I exposed it so that callers too could get the decompressed Image header without decompressing the HDU.  I didn't see a better way to do that.  I'll rename it though, and I neglected to show its usage in the `README`.

at88mph:
Thanks @attipaci .  Rework submitted.

src/main/java/nom/tam/fits/compression/algorithm/hcompress/HCompressorQuantizeOption.java
```diff
@@ -32,13 +32,18 @@
...
 import nom.tam.fits.compression.algorithm.quant.QuantizeOption;
+import nom.tam.fits.compression.algorithm.rice.RiceCompressOption;
 
 public class HCompressorQuantizeOption extends QuantizeOption {
 
     public HCompressorQuantizeOption() {
         super(new HCompressorOption());
     }
 
+    public HCompressorQuantizeOption(RiceCompressOption compressOption) {
```
attipaci, line 43: Here the argument needs to be `HCompressorOption`, to Rice... (I also guess this was one of the lines not covered in the tests...)

attipaci (changes requested):
It looks like there was just one small copy/paste issue in the latest changes that needs to be attended. There is also a (small) drop in the coverage -- what looks like 2 lines. Unfortunately Codecov is in a funk again (seems to happen a lot) so I cannot be sure where the coverage change originates from, although I suspect one line is to do with the above.... If you cannot find the other, no big deal -- we can live with a 0.01% drop in coverage (and it will pass on master).

src/main/java/nom/tam/fits/compression/algorithm/hcompress/HCompressorQuantizeOption.java
at88mph: Working on this now, thanks.

attipaci (approved):
Despite the small coverage decrease, I'll give it a pass and merge...

PR #408, opened 2023-05-15
Cache randomValues for dithering

### LINKED ISSUE #407: Cache randomValues for dithering
We've been looking at speeding up the reading of FITS files, and found what seems to be an easy speedboost.

`QuantizeProcessor.DitherFilter#initRandoms` is run for every DitherFIlter creation on every worker thread. These values don't seem to change, and from some testing making the field static has no effect on the resulting image, but has improved the opening time. For a stack of 100 LCO images is saves about 2 seconds on my machine.

Wanted to discuss this change with with you to see if there are any traps with making the field static, or otherwise sharing the the result.

keastrid:
Fixes #407

### DIFF
--- a/src/main/java/nom/tam/fits/compression/algorithm/quant/QuantizeProcessor.java
+++ b/src/main/java/nom/tam/fits/compression/algorithm/quant/QuantizeProcessor.java
@@ -154,15 +154,14 @@ public class QuantizeProcessor {
 
         private static final double RANDOM_START_VALUE = 16807.0;
 
+        private static final double[] RANDOM_VALUES = initRandoms();
+
         private int iseed = 0;
 
         private int nextRandom = 0;
 
-        private final double[] randomValues;
-
         DitherFilter(long seed) {
             super(null);
-            this.randomValues = initRandoms();
             initialize(seed);
         }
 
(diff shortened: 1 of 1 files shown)


attipaci:
Hi Kevin!

Unfortunately it's not going to be quite that easy. Here's how to make this PR work, in my opinion:

1. First you want to move the __ private static__ `RANDOM_VALUES` and the __private static__ `initRandoms()` methods up to the top level `QunatizeProcessor`, since the non-static inner classes cannot have static fields. That means also moving the related constants from `DitherFilter` along.... This will also require chaning the references of `this.randomValues` in `DitherFilter` to just `randomValues` (without `this` prepended).

2. SEcondly, you probably want to delay initialization until a `DitherFilter` is actually used so as to not create unnecessary computation when it's not needed. That means, you don't want to initialize `RANDOM_VALUES` as you declare it, but only when `DitherFilter` needs it. In practice, that will require two things:

     2.1. Remove the __final__ modifier from `RANDOM_VALUES`, and as a result refactor it back to `randomValues` as before (still keeping the __private static__ modifiers however).

    2.2. Initialize the random sequence in a thread-safe way in `DitherFilter.initialize()`, e.g. by inserting the following up front:

     ```java
       synchronized (randomValues) {
             if (randomValues == null) {
                  randomValues = initRandoms();
             }
        }
     ```
I tried it myself and it passed the build and tests all fine, but I'll let you fix up your PR unless you rather just have me do it...

keastrid:
Hi Atila,

I've pushed the requested changes. We shouldn't be synchronizing on a non-final field though. 

It should be safe as a final value, as the static body doesn't get run until the class is initialized.

PR #411, opened 2023-05-16
Issue 407: Calculate dither random sequence only once

attipaci:
when needed, and faster than before...

Supplants PR #408.


attipaci:
@keastrid,

Sorry to supplant your PR, but it became more involved than I originally expected, and it was simply easier to create this new PR rather than trying to fix up yours. I hope you can check it out and verify that it does for you what you need. I'll wait for you to confirm before I merge to master, and produce an rc3 sometime soon after...

Thanks once again!

keastrid:
I don't mind at all.

This looks good, the values are only being made once, and is a very nice performance improvement, thanks!

PR #414, opened 2023-05-21
Speed up array IO  (2.5x - 4x) with nio views as suggested by Kevin Eastridge

attipaci:
@keastrid has noticed that array IO can be significantly improved by using `java.nio` views (see Discussion #413), so this PR implements that, and provides some minor fixes/improvements to the FITS IO implementation. The buffer views are used for the retrieval of multiple elements only, while single elements are retrieved more efficiently using, e.g. `ByteBuffer.getFloat()` directly.


keastrid:
This works very well, thank you! It is a very noticeable improvement in opening times.

I see you added the same for fits writing, which is a great addition.

attipaci:
More work was done (above) to:

- Make the code more elegant (less repeated)
- Retain image read performance for arrays where trailing dimension may be small.
- Extend high-throughput reads to Random Groups and Binary Tables also
- Fill missing documentation

Will need to fill new test coverage holes before merging...

attipaci:
I did some benchmarking of the new IO speed vs previous releases. In general, reading and writing images is 2.5x -- 4x faster than previous version of the library (e.g. 1.15.2 to 1.17.1) on my laptop with a Gen 3 NVMe SSD. And, reading from `FitsInputStream` (aka `BufferedDataInputStream` in earlier versions) is a whopping 20-25 times faster than 1.15.2...

keastrid:
@attipaci I think the recent commits introduced a regression to table reading.

Trying to get the hdu array results in the following exception:
```
May 22, 2023 9:07:47 PM nom.tam.fits.Header forceEOF
WARNING: Junk detected at -1.
nom.tam.fits.FitsException: Not a proper FITS header:  at -1
	at nom.tam.fits.Header.checkFirstCard(Header.java:2124)
	at nom.tam.fits.Header.read(Header.java:1579)
	at nom.tam.fits.Header.readHeader(Header.java:170)
	at nom.tam.fits.Fits.readHDU(Fits.java:1005)
	at nom.tam.fits.Fits.readToEnd(Fits.java:1045)
	at nom.tam.fits.Fits.read(Fits.java:956)
```

It is also only finding 2 of the 4 hdus. The file is available [here](https://mast.stsci.edu/api/v0.1/Download/file/?uri=mast:TESS/product/tess2021175071901-s0040-0000000008266786-0211-s_tp.fits)

attipaci:
I was fearing something like this would happen, even though the built-in regression tests passed. Based on the error, it's the binary table (2nd HDU) that reads either too short or beyond its boundary, so it's trying to detect the next HDUs header at the wrong offset -- hence without success. I'll see what I can make of it and create a fix... And, I'll create a new issue for tracking the regression...

PR #421, opened 2023-05-23
Fix duplicate skip padding after binary table data segment

attipaci:
A bug in PR #396 resulted in having skipped the amount of padding after binary tables twice when reading from streams, and hence losing alignment to the beginning of the following HDU. The problem did not affect reading from files.


attipaci:
The decrease in coverage is dies to lines removed. The same lines are covered as before. Hence, OK to merge...

PR #424, opened 2023-05-24
Code cleanup

attipaci:
Make the code more readable by imposing a degree of uniformity:

- remove empty trailing spaces and multiple empty lines
- format code to the same standard line width across the package
- organize imports consistently
- remove unnecessary `this` and `super` markers (these don't serve a purpose other than providing a visual clue to programmers. However most moder code editors will use highlighting to distinguish variables of different scope much more effectively -- and so the use of these prefixes is more a distraction than a useful feature at this point. You are welcome to disagree...)
- Use increment `++` and decrement `--` operators instead of ` += 1` and ` -= 1`.


attipaci:
The slight decrease in coverage is due to code removed (i.e. shortened code base) rather than missing new lines of code. As such, I'll give it a pass.

PR #442, opened 2023-06-09
Simple compressed tiles

attipaci:
Provide ways to make accessing select areas of a compressed image or table more accessible.

This PR also contains a complete overhaul of `ColumnTable`


attipaci:
Much of the wish-list is now implemented. Next is to get coverage back up... It will take a few days for sure...

attipaci:
@at88mph and @keastrid -- This PR is nearly ready for showtime. I still need to add a few tests for the new simplified compressed tile access, but what was there before should be working. If you have time to test it, you might try this branch now. I'll aim to complete the test coverage next week, so if you find new issues, I can still work on them before merging towards the end of next week.

Most things should work as before (only faster and better). There are some corner cases in `BinaryTable` that behave differently from before (nagging inconsistencies that have been reconciled). But, hopefully you either won't encounter these, or else find an easy way around... (Feel free to look at the open issues to get an idea what changes are coming -- and how these might affect, or help, you...)

PR #459, opened 2023-07-12
Common Data padding implementation.

attipaci:
Each `Data` subclass was implementing its own padding in `write()`, but really it's safer and better to have a common implementation in `Data` itself.  `Data.write()` now calls `.writeUnpadded()` and then adds padding with the byte value specified by `.getPaddingByte()` by default.

The changes will affects external `Data` implementations, which must override `writeUnpadded()` instead of `write()` as a result. However, it is very unlikely that anyone would need to extend `Data` beyond the subclasses provided by this package, so we go ahead with the change noting the very slight risk of breaking some niche applications. In the unlikely case someone did implement `Data` externally, the fix should be as easy as renaming the overridden `write()` method to `writeUnpadded()` to fix.

### DIFF
--- a/src/main/java/nom/tam/fits/AsciiTable.java
+++ b/src/main/java/nom/tam/fits/AsciiTable.java
@@ -922,7 +922,7 @@ public class AsciiTable extends AbstractTableData {
     }
 
     @Override
-    public void write(ArrayDataOutput str) throws FitsException {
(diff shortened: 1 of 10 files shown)


attipaci:
Will not merge, because while it provides a little more consistency in implementation, it does it at the cost of breaking external `Data` implemetation, which though unlikely are nevertheless possible. It's just not worth meddling in this case.

PR #461, opened 2023-07-13
Issue 460 -- Rice compression BYTEPIX regression in 1.18.0-rc6

attipaci:
The BYTEPIX value was not properly shared across copies of `RiceCompressOption`. Fix by shared config properties in `RiceCompressOption` and also `HCompressorOption` classes. Also validate BYTEPIX and BLOCKSIZE settings (in `RiceCompressOption`) and scale parameter (in `HCompressorOption`) and throw an `IllegalArgumentException` if the set
values are invalid.


attipaci:
Will add tests for added compression option settings validation...

keastrid:
This appears to work on our end, thanks!

PR #467, opened 2023-07-26
Fix ASCII table read regression

attipaci:
ASCII tables can only be read after `FitsFactory.setUseAsciiTables(true)` was set prior. This is a regression in 1.18.0.


attipaci:
I will merge after adding an appropriate test case for preventing this type of regression in the future....

PR #478, opened 2023-08-20
Issue 476 -- Checksum verification on stream inputs

attipaci:
added automatic checksum aggregation to `FitsInputStream`, and changed HDU / data read methods to store the accumulated stream checksums for verification. The extra calculation does not seem to impact performance (at least in my benchmarks), consistent with reads being IO-limited, not CPU-limited typically. (My tests were performed with a 7th gen Intel Core i7 CPU and a Gen3 nVME SSD).


attipaci:
Still need test cases to prevent decreased coverage...

attipaci:
Coverage restored. Ready to merge...

PR #491, opened 2023-09-18
Targeted Data creation, improved RandomGroups and UndefinedData support

attipaci:
- Introduced `ImageData.from(Object)`, essentially offering the same behavior as the constructor with the same argument. Both now check that the argument is a valid regular primitive numerical Java array, or else throw an `IllegalArgumentException`.
- Added `Data.toHDU()` implementations, to wrap data classes into HDUs with the mandatory minimal header descriptions.
- Inspection methods for `UnknownData`.
- Header order changed s.t. EXTEND comes after GCOUNT, resulting in better conformance to the FITS standard. (EXTEND is an optional keyword. As such it must come after the mandatory keywords.)
- `Data.fillHeader()` implementations to remove SIMPLE + EXTEND or XTENSION keywords depending on which types of HDUs the data may support (`ImageData` will assume primary HDU until it becomes clear it must be written as an extension).
- The FITS standard allows random groups in the primary HDU only. As such we'll throw an exception if trying to write random groups data as extensions.
- New methods to access named parameters and individual images in random groups data

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -358,7 +358,6 @@
           <configLocation>${project.basedir}/src/main/checkstyle/nom-tam-fits-style.xml</configLocation>
           <suppressionsLocation>${project.basedir}/src/main/checkstyle/checkstyle-suppressions.xml</suppressionsLocation>
           <suppressionsFileExpression>checkstyle.suppressions.file</suppressionsFileExpression>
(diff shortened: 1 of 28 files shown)


attipaci:
Need to add tests to restore coverage.

PR #496, opened 2023-10-19
Work around Windows issue with read-only files on network shares.

### LINKED ISSUE #495: Exception thrown creating Fits objects from read-only files on network share
On Windows, when attempting to create a ``Fits`` object from a path to a read-only file, an exception can incorrectly be thrown if that path is on a network share.

This is due to the fact that ``File::canWrite()``, as used by ``Fits::randomInit()``, can incorrectly return ``true`` for read-only files on network shares when running on Windows. When that happens, ``Fits::randomInit()`` attempts to open the file in ``rw`` mode, which throws an exception because the file is not actually writable.

The issue seems to be limited to Windows.

Because, for example, ``Files::isWritable()`` has similar issues, it might be simplest to first attempt to open the file in ``rw`` mode, and if that fails, fall back to read-only. An MR that takes this approach to address the issue is incoming.

cek:
This MR resolves #495.

On Windows, ``File::canWrite()`` can incorrectly return ``true`` for read-only files on network shares. This can cause ``Fits::randomInit()`` to attempt to open the read-only file in ``rw`` mode, which throws an exception because the file is not actually writable.

This MR modifies ``Fits::randomInit()`` to first attempt to open the named file in ``rw`` mode. If that fails, it then attempts to open it in read-only mode, as before.

### DIFF
--- a/src/main/java/nom/tam/fits/Fits.java
+++ b/src/main/java/nom/tam/fits/Fits.java
@@ -908,18 +908,21 @@ public class Fits implements Closeable {
     @Deprecated
     protected void randomInit(File file) throws FitsException {
 
-        String permissions = "r";
         if (!file.exists() || !file.canRead()) {
             throw new FitsException("Non-existent or unreadable file");
         }
-        if (file.canWrite()) {
-            permissions += "w";
-        }
         try {
-            dataStr = new FitsFile(file, permissions);
+            // Attempt to open the file for reading and writing.
+            dataStr = new FitsFile(file, "rw");
             ((FitsFile) dataStr).seek(0);
         } catch (IOException e) {
-            throw new FitsException("Unable to open file " + file.getPath(), e);
+            try {
+                // If that fails, try read-only.
+                dataStr = new FitsFile(file, "r");
+                ((FitsFile) dataStr).seek(0);
+            } catch (IOException e2) {
+                throw new FitsException("Unable to open file " + file.getPath(), e2);
+            }
         }
     }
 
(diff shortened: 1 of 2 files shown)


attipaci (approved):

attipaci:
Hi @cek,

Thanks for the fix. It looks good. Before I merge, can you add unit tests for the new lines of code? Ideally, you should have 100% diff coverage unless there is essential unreachable code (which should not be the case here). Once you have the test coverage, I'll be happy to merge it. 

Thanks in advance. :-)

cek:
Sure thing, apologies.

Should the test be a standalone under ``src/test/java/nom/tam/fits``, or ``src/test/java/nom/tam/fits/test``, or added to an existing test module?

Similarly, should the associated ``.fits`` file be added to the ``blackbox-images`` repo, or ``src/test/resources``?

attipaci:
You can add the tests to an existing test module -- as you see fit.

I'm not sure you need to add a FITS at all. In the test you can just generate one ad-hoc (typically under `target/`) and then set permissions on it to pass/fail the relevant parts of the open routine. If you think there is a strong reason why you should use a specific external test file, you can put it under `src/test/resources/`, e.g. `src/test/resources/nom/tam/fits/` if it's small. And, if it's large, then it would go to blackbox-images...

I hope that helps.

cek:
You're right, of course, the test can simply generate a stub fits.

Sorry this took a couple of iterations; let me know if I've missed anything about the CI environment or the like that might cause (more) issues.

attipaci:
@cek,

Thanks. It looks OK to me, don't worry -- I stumble around more than that ;-). I just re-ran the CI once more because Codecov did not pick up the coverage report after the last change (it happens sometimes). It looks good. I'll go ahead an merge it. Thanks again!

PR #517, opened 2023-12-12
Fix constructor throw vulnerabilities by declaring finalize() final.

attipaci:
The latest maven spotbug plugin exposed a bunch of vulnerabilities with classes that may throw an exception from their constructors. In principle a program may extend these classes, and override the `finalize()` method to expoit the partially initialized state of the classes involved. In the context of the the nom-tam-fits library there is no sensitive data that could be exposed in this way, so it's not really a vilnerability in the true sense.

Also, the easiest fix for it would be to declare the `finalize()` method of these classes `final` in order to prevent intercepting partially initialized data. However the `finalize()` method has been deprecated since Java 9. Thus, that approach to tending to the issue is not future proof itself.

We'll just keep the Spotbug detectors for this type of problem disabled. Some related fixes are in PR #518.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -239,7 +239,6 @@
         <version>3.6.3</version>
         <configuration>
           <source>8</source>
-          <aggregate>false</aggregate>
           <failOnError>true</failOnError>
           <doclint>all,-missing</doclint>
         </configuration>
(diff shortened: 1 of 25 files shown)


attipaci:
See PR summary on why this will not be merged.

PR #552, opened 2024-02-15
Eliminate ColumnDesc.fileSize in favor of .rowLen()

attipaci:
`ColumnDesc.fileSize` which specified the size of column entries in the FITS was not maintained consistently. It is better to use `ColumnDesc.rowLen()` instead to calculate the size of entries on an as-needed basis.

### DIFF
--- a/.github/workflows/maven.yml
+++ b/.github/workflows/maven.yml
@@ -53,7 +53,7 @@ jobs:
         echo "branch coverage = ${{ steps.jacoco.outputs.branches }}"
        
     - name: Upload coverage to Codecov
-      uses: codecov/codecov-action@v3
+      uses: codecov/codecov-action@v4
       continue-on-error: true
       with:
         working-directory: ./nom-tam-fits
@@ -62,7 +62,7 @@ jobs:
         flags: unittests
         name: codecov
         verbose: true
-#        token: ${{ secrets.CODECOV_TOKEN }}
+        token: ${{ secrets.CODECOV_TOKEN }}
         
     - name: Upload coverage to Coveralls.io
       continue-on-error: true
--- a/.github/workflows/maven.yml
+++ b/.github/workflows/maven.yml
@@ -53,7 +53,7 @@ jobs:
         echo "branch coverage = ${{ steps.jacoco.outputs.branches }}"
        
     - name: Upload coverage to Codecov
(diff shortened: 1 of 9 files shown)


keastrid:
Thanks for the quick fix!

The output table from the test case appears to work, and trying it with a larger table of actual data also seems to work.

Incremental writing of fits files also seems to be working.

funpack will decompress the example table, but outputs a table with only the first row having a value of 1, the rest are 0. Nom.tam will read it without issue, so I am not sure what is causing this. I have attached them
[here](https://github.com/nom-tam-fits/nom-tam-fits/files/14316215/Test.zip).

attipaci:
@keastrid, Please check out this PR to see if it fixes the compressed table readability issue. (I'll do my own testing in the coming days also).

attipaci:
@keastrid, Cool, I'm happy this worked. (Less happy that such a nasty bug sneaked in, and went unnoticed so long.) I'll check the compressed header again more carefully. Perhaps some other header value is also incorrect, causing funpack to fail (while nom-tam-fits may work with some default, or perhaps because of some non-standard but consistent mishandling...)

I'll also want to see how long this bug existed. I may have introduced it as a regression in 1.18.0, when I have overhauled binary tables, or it may have existed long before then. In either case it would be good to know what releases were affected and put a notice on this bug accordingly...

keastrid:
Sounds good!

attipaci:
Hi @keastrid,

I looked more carefully into the compressed table generated by your example code, and also went trough the compressed table specification in the FITS 4.0 document.

From what I can tell nom-tam stores the table properly. In your example 8 scalar double-precision (1D) columns are compressed into a single tile using `GZIP_2`. They are then stored on the heap as a variable byte array of 34 bytes, starting at offset 0. I'm not sure why funpack is not processing it right, short of the GZIP_2 implementation being incorrect (maybe in nom-tam).

The nom-tam `GZIP_2` uses the standard Java `GZIPOutputStream`, except that it performs the shuffling of bytes before compressing (see `nom.tam.fits.compression.algorithm.gzip2.GZip2Compressor.shuffle()`). Perhaps there is a bug in how the shuffling is performed, which would then result in garbage when done differently, e.g. by funpack... I cannot think of much else wrong here...

PR #553, opened 2024-02-16
Header.findCards() with test added

rmathar:
Adds a function in the Header class that searches for all cards in a header that match a regular expression. Associated test function added to fits/HeaderTest.java scanning one of the existing simple fits images.

### DIFF
--- a/.github/workflows/maven.yml
+++ b/.github/workflows/maven.yml
@@ -64,6 +64,7 @@ jobs:
         verbose: true
 #        token: ${{ secrets.CODECOV_TOKEN }}
         
+
     - name: Upload coverage to Coveralls.io
       continue-on-error: true
       env:


attipaci:
Hi @rmathar!

Thanks for preparing the PR and the tests. Unfortunately, the codecov-actions update 4 days ago has broken the coverage upload (not just for us, of course). I have not reverted that on master, but e won't get coverage reports here until you rebase this PR to the current head on master... Could you do the rebase and a force push, if it's not too much trouble?

Alternatively, I can check the coverage locally (which is certainly OK if you do not want to deal with the rebase). I just want to be sure that all your new code is sufficiently covered by the tests. I'll also review your code just in case I have some comments or suggestions. But I expect that I'll merge this very soon, and include it in the upcoming 1.19.1 release, expected around March 15...

PR #558, opened 2024-02-18
Header.findCards() with test added

rmathar:
closes https://github.com/nom-tam-fits/nom-tam-fits/discussions/546 if pulled

### DIFF
--- a/src/main/java/nom/tam/fits/Header.java
+++ b/src/main/java/nom/tam/fits/Header.java
@@ -11,6 +11,7 @@ import java.util.Comparator;
 import java.util.HashSet;
 import java.util.List;
 import java.util.Set;
+import java.util.Vector;
 import java.util.logging.Level;
 import java.util.logging.Logger;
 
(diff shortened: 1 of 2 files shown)


src/main/java/nom/tam/fits/Header.java
```diff
@@ -908,6 +909,43 @@ public HeaderCard findCard(String key) {
...
+    * some HIERARCH string conventions of unspecified depth. So to search for
+    * some key like "HIERARCH OBO SUBOBO MYOBO", which would appear with the
+    * key HIERARCH.OBO.SUBOBO.MYOBO in this FITS implementation, one could
+    * search with regex="HIER.*MYOBO" and find it, supposed FitsFactory.setUseHierarch(true)
+    * was called before creating the header.
+    * 
+    * @param regex The generalized regular expression for the keyword search
+    * @return The list of header cards that match the regular expression.
+    * @author Richard J. Mathar
+    * @since 2015-03-03
```
attipaci, line 925: `@since` is not correct for the library. I recommend you change it to `@since 1.19.1`.

src/main/java/nom/tam/fits/Header.java
```diff
@@ -908,6 +909,43 @@ public HeaderCard findCard(String key) {
...
+    */
+    public HeaderCard[] findCards(final String regex) { 
+        /* The collection of header cards that match.
+        */
+        Vector<HeaderCard> crds = new Vector<HeaderCard>();
```
attipaci, line 930: I think `ArrayList` instead of `Vector` here since there is no chance of concurrent access here, and so the lighter-weight `ArrayList` should be faster / less resource hungry...

attipaci (changes requested):
Looks good. Just two minor requests from me, and I'll be ready to merge it.

src/main/java/nom/tam/fits/Header.java
rmathar: Done in my branch. Do I need to delete the old PR and generate a new one?

attipaci (approved):

Issue #567, opened 2024-03-09
Table compression results in unpredictable heap ordering


attipaci:
Tile compression is parallelized. Because there is no specific order in which parallel tiles produce compressed data, there is no predictable order in which tiles put data on the heap. While the compressed data is itself valid, the irreproducible behavior is a nuissance. We should therefore defragment the compressed heaps after the compression of tiles, to produce a predictable, and also more effecient layout for reading.

Issue #563, opened 2024-03-01
Allow RICE_1 and HCompress with defaults (also Rice for tables)


attipaci:
Our constructors for Rice / HCompress compressors require a non-null option argument, which in practice means that users must supply options explicitly when using the algorithm to compress. However, FITS defines default values for the options, so we should allow using `RICE_1` or HCompress without providing them options explicitly. This is also necessary to enable table compression with `RICE_1`, which we currently do not support for this very reason.

In practice that means that currently only the optionless `GZIP_1`, `GZIP_2` algorithms can be used for tables.

On a related note, we do allow other 'optionless' algorithms on tables, such as the lossy PLIO, which is not allowed for tables as per FITS 4.0. So, we should also check and throw an appropriate exception is someone tries to compress binary tables with an unapproved algorithm.

keastrid:
I'm not sure all of these are possible currently, as per the FITS 4.0 spec, p. 49:

> The permitted algorithms for compressing BINTABLE columns
are ’RICE 1’, ’GZIP 1’, and ’GZIP 2’ (plus ’NOCOMPRESS’),
which are lossless and are described in Sect. 10.4. Lossy compression could be allowed in the future once a process is defined
to preserve the details of the compression.

attipaci:
OK, but currently we do not allow `RICE_1` from that list for the reasons mentioned above. So the issue is valid. And maybe more so, because I don't think we have anything preventing us from using the lossy `PLIO` at the same time...

attipaci:
I'll change the ticket description somewhat...

keastrid:
Yes, the issue is still valid. 

The compression options should definitely be limited to what is actually available for tables.

attipaci:
OK, in the PR in prep, `RICE_1` now works with 8 -- 32-bit integers (the standard allows for 64-bit also, but there is currently no algorithm anywhere I know of that supports it -- cfitsio certainly does not).

And `CompressedTableHDU.fromBinaryTable()` will throw an `IllegalArgumentException` if supplied with an algorithm that is not suitable for table compression...

Issue #562, opened 2024-03-01
Compressed table HDU ZCTYPn keyword written with 0-based index


attipaci:
The `ZCTYPn` keyword was written into the header of compressed table HDUs with a 0-based (Java) index, instead of the 1-based indexing expected by FITS. On reading the indexing was correct, but in any case this bug rendered compressed tables unreadable both by us and by other tools.

Will fix for 1.19.1.

Issue #557, opened 2024-02-17
Compression of variable-length arrays (VLAs) in binary tables


attipaci:
We don't currently implement the compression of VLAs in binary tables according to specification. While we compress the heap pointers, we leave the heap data uncompressed. But that is not the standard that we are supposed to follow. Instead the FITS standard 4.0 describes ah method of compression both the heap, and the pointers (64-bit Q-type pointers for the compressed heap, and P or Q-type pointers for the uncompressed heap).

We therefore should implement the compression of VLAs to provide more complete support of binary table compression.

__UPDATE__:

As it turns out the VLA compression implemented by CFITSIO and its `fpack` does not actually follow the FITS 4.0 standard (or the original Pence+2013) convention. It is likely, that the maintainers of the standard and CFITSIO will choose to update the standard to match the actual behavior. For this reason, we shall provide VLA compression that is CFITSIO compatible by default. However, because there is a chance that one may need to read compressed VLAs produced by other tools, which actually followed the original prescription, we also provide `CompressedTableHDU.useOldStandardVLAIndexing(boolean)` to allow selecting the convention to use for decompressing tables with VLA columns.

attipaci:
This is actually a pretty nasty thing to fix. All the binary table compression was implemented to work not on the `BinaryTable` data but on the regular 2D `ColumnTable` element of it. As such the compression has no sense of the heap whatsoever, and therefore cannot possibly operate on the heap of the (unknown) encompassing `BinaryTable`.

I can think of two ways to 'fix' this:

 1. The clean way is to reimplement the compression classes to work with `BinaryTable` instead of the dumb `ColumnTable` component. This would keep the API simpler, and similar, but would break backwards compatibility on functions that likelyno-one ever used. (These functions should not have been exposed to users in the first place!)

 2. Keep API compatibility, and add work around by adding a bunch of bolt-on functions, which must be used together to allow the table compression classes to access the heap data.

I'm debating which is the less bad way to proceed from this whole that was dug when compression was added in this form....

attipaci:
Hi @keastrid,

I think I'm well on the way to sorting out the binary table compression issues. I have now modified the table tile compression/decompression classes to add a proper procedure for compressing/decompressing variable-sized columns. I have verified that this works more or less how it is is described in the FITS standard. I have checked the storage of compressed table entries the compression/decompression of the compressed and original heap pointers, and the recovery of the original data through them.

[...]

PR #568, opened 2024-03-10
Table compression fixes

attipaci:
Several fixes to table compression.

- #557
- #562
- #563
- #567
- #571 
- #575
- #576

### DIFF
--- a/.github/workflows/maven.yml
+++ b/.github/workflows/maven.yml
@@ -77,8 +77,8 @@ jobs:
         fi
       working-directory: ./nom-tam-fits
         
-    - name: Check ppackage
-      run: mvn package -Ppublic-api
+    - name: Check package
+      run: mvn clean package -Ppublic-api
       working-directory: ./nom-tam-fits
         
   release:
--- a/README.md
+++ b/README.md
@@ -1799,11 +1799,15 @@ you should be aware of some potentially severe pitfalls.
  1. VLA table compression is not widely supported by tools (including CFITSIO's own `fpack` tool! -- see more on
  that below).
 
- 2. Second, the [(C)FITSIO](https://heasarc.gsfc.nasa.gov/fitsio/) implementation diverges from the documented 
+ 2. The [(C)FITSIO](https://heasarc.gsfc.nasa.gov/fitsio/) implementation diverges from the documented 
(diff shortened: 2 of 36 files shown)


attipaci:
@keastrid,

This PR is ready on my end, now with as much test coverage as can be added without contorting oneself. 

I did confirm interoperability with `fpack`, within the limits that the (ever multiplying) bugs in `fpack` allow. But, I'll wait until your checks before merging it to master, and preparing an release candidate with it...

keastrid:
Thanks! 

It is unfortunate the CFITSIO has had so many issues.

The round trip of data to compressed table and back again via nom.tam seems to work.

I gave it a quick test using the same code as #550, and when trying to funpack it I get:
```
Error processing file: test.fits.fz
  in HDU number 2

FITSIO status = 414: error uncompressing image
Could not find the required ZPCOUNT keyword
Input file is unchanged.
```

Do you see the same?

attipaci:
Hi @keastrid,

I'll look again. Perhaps something I changed during cleaning up because I did not see that error before, at least not with the fpack I have on Fedora...

keastrid:
I was also using the same fpack version, shipped via ubuntu (via wsl).

This is with incremental writing, your recent changes seem to have fixed the issue with the test case. I call `toHDU` then set the column names.

When trying to work with compressing tables with a string column (apparently funpack doesn't allow that), I was trying to disable compression for that column by passing `NO_COMPRESS` as indicated by the javadoc on `CompressedTableHDU#fromBinaryTableHDU`; however, the binary tables only allow the keyword `NOCOMPRESS`, which is not recognized by funpack... 

`NOCOMPRESS` is specified in the FITS spec, so this seems like another fpack bug as well as a documentation typo in nom.tam.

keastrid:
You can pass it along since you already have a line of communication open with them.

Sounds good on the release schedule, thanks!

attipaci:
@keastrid, Thanks for all your help in resolving this 'can-of-worms'. It's really good that you are pushing what this library can do, and I am also happy that the giant mess of how compression was added to this library is slowly getting sorted, at least to the point where it's starting to work more or less as expected, even if the code underneath remains a ridiculously entangled mess...

Issue #636, opened 2024-06-20
Compressed tables lose column names


keastrid:
Just updated to 1.20 (we had previously just pulled in the compression fix), and we found an issue when compressing tables - they lose the column names and take on the new default column names

This code is enough to reproduce it:
```java
var fits = new Fits();
var table = new BinaryTable();
table.addColumn(new double[]{1, 2, 3, 4});
var h = table.toHDU();
h.setColumnName(0, "test", "test");
var compressedHdu = CompressedTableHDU.fromBinaryTableHDU(h, -1);
compressedHdu.compress();
h = compressedHdu;
fits.addHDU(h);
fits.write(Path.of("test.fits").toFile());
fits.close();
```

Commenting out `h = compressedHdu;` produces a table with the correct headings.

attipaci:
Ouch! It's too bad you did not catch this in the weeks (months) before the final release. I'll look into finding a fix for it...

keastrid:
Sorry, it was on my todo then other things happened. Though I'm not sure I would've caught it quickly.

attipaci:
No worries... :-D

PR #637, opened 2024-06-20
Issue #636: Keep column names in compressed tables.

attipaci:
Do not set default column names when creating compressed tables


attipaci:
@keastrid, I think this PR should cure the issue with the default column names in compressed tables. Let me know if it works for you or if you find other problems.

keastrid:
This seems to work, thanks!

PR #665, opened 2024-08-21
Recalculate stored checksums when writing headers

attipaci:
Also:

- Mandatory keyword comments are retained as in the original FITS
- `CHECKSUM` and `DATASUM` keywords now have timestamps in comment on when they were last updated. This is best practice that is recommended by the checksuming convention.
- Deprecated `FitsChecksum.checksum(...)` methods, as they aren't useful to users. Instead users probably want to call `BasicHDU.verifyIntegrity()` to verify checksums.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -1,3 +1,4 @@
+
 name: Build
 
 on: 
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -1,4 +1,3 @@
-
 name: Build
 
 on: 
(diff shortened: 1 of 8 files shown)


attipaci:
@johnastromurphy, I think this PR fixes the issue with `setChecksum()` now working as expected because the header was modified during the `write()`. See if you find more issues or corner cases. I might tweak the solution a little more too to make it more foolproof...

johnmurphyastro:
Hi @attipaci 
Yes, that works great. That was quick!
Regards, John Murphy

PR #697, opened 2024-11-04
Revised calculations of space available in cards

attipaci:
Match the calculation of space available for values in HIERARCH cards to the flexible assignment string used by PR #696. Also revise space calculations in general, to allow for maximal used of header cards space given the value constraints and FITS settings.


attipaci:
@vforchi, as expected the flexible spaces around the '=' are more complicated than it seemed at first. This PR is meant to clean up the fallout from #696...

It looks like it works, but I'll have to cover some new lines with tests before I can be more sure...

PR #717, opened 2024-12-20
feat: added file channel implementation for read and write of random …

at88mph:
…access

### DIFF
--- a/src/main/java/nom/tam/util/RandomChannelFileIO.java
+++ b/src/main/java/nom/tam/util/RandomChannelFileIO.java
@@ -90,17 +90,17 @@ public class RandomChannelFileIO implements RandomAccessFileIO {
 
     @Override
     public FileDescriptor getFD() {
-        throw new UnsupportedOperationException("FileDescriptor is not supported for FileChannels.");
+        throw new UnsupportedOperationException("FileDescriptor is not implemented for FileChannels.");
     }
 
     @Override
     public void setLength(long l) {
-        throw new UnsupportedOperationException("Setting file length is not supported for FileChannels.");
+        throw new UnsupportedOperationException("Setting file length is not implemented for FileChannels.");
     }
 
     @Override
     public void writeUTF(String s) {
-        throw new UnsupportedOperationException("Write operation is not implemented for FileChannels.");
+        throw new UnsupportedOperationException("writeUTF operation is not implemented for FileChannels.");
     }
 
     @Override
@@ -128,7 +128,7 @@ public class RandomChannelFileIO implements RandomAccessFileIO {
 
     @Override
     public int read() {
-        throw new UnsupportedOperationException("read()");
+        throw new UnsupportedOperationException("read is not implemented for FileChannels.");
     }
 
     /**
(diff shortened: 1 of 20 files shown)


attipaci:
Hi @at88mph,

I'll check the PR in the New Year. Until then, happy holidays to you!

at88mph:
To you as well, @attipaci !

attipaci (commented):
Can we change `RandomChannelFileIO` to `RandomFileIOChannel`? After all it's a channel that does random(-access) file I/O...

src/test/java/nom/tam/util/RandomChannelFileIOTest.java
```diff
@@ -0,0 +1,220 @@
...
+
+import java.io.File;
+import java.io.FileWriter;
+import java.nio.charset.StandardCharsets;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.nio.file.Paths;
+
+
+public class RandomChannelFileIOTest {
```
attipaci, line 46: Test name to match changed class name, i.e. `RandomFileIOChannelTest`

src/main/java/nom/tam/util/RandomChannelFileIO.java
```diff
@@ -0,0 +1,175 @@
...
+import java.util.logging.Logger;
+
+/**
+ * Concrete implementation of RandomAccessFileIO for FileChannels.  This is a generic implementation using the Java
+ * FileChannel API.
+ * An example of an implementation of the underlying FileChannel is the NIO SPI file system provider to Amazon S3.
+ *
+ * @see <a href="https://github.com/awslabs/aws-java-nio-spi-for-s3">AWS NIO SPI library</a>
+ */
+public class RandomChannelFileIO implements RandomAccessFileIO {
```
attipaci, line 53: I think `RandomFileIOChannel` is a better name. After all it is a Channel that does Random(-access) File I/O.

src/main/java/nom/tam/util/RandomChannelFileIO.java
```diff
@@ -0,0 +1,175 @@
...
+
+    @Override
+    public void writeUTF(String s) {
+        throw new UnsupportedOperationException("writeUTF operation is not implemented for FileChannels.");
+    }
+
+    @Override
+    public void close() throws IOException {
+        // Just in case (i.e. this is a unit test running), check for null.
+        if (fileChannel != null) {
```
attipaci, line 109: I think by the public constructor `fileChannel` cannot be `null`, so no need to check here. Of course, one could in principle concoct a way to access the package-level constructor also, and pass `null` there to initialize `fileChannel`. Maybe the testing constructor could throw an `IllegalArgumentException` if the `fileChannel` argument is `null`.

attipaci (commented):
Overall it looks good. I just have a couple nitpicking requests for small changes.

attipaci:
@at88mph Sorry it took me this long to get to it. I was tied up with a whole bunch of other projects, so this got pushed back more than I intended to. The PR looks pretty good. Thanks again for prepping it!

at88mph:
Done.  Thank you @attipaci.

attipaci (approved):

PR #723, opened 2025-02-19
Javadoc edits for newly added `RandomFileIOChannel`

attipaci:


at88mph (approved):

attipaci:
@att88mph,

Sorry, I could not resists changing the class name again. I was unhappy with it because it sounded so awkward... I slept on it, and decided that `RandomAccessFileChannel` is a bit more digestible. I also declared methods that return actual values as `final` since no subclasses should ever mess with these. See if you approve.

PR #865, opened 2026-07-02
Compress cutout fix

at88mph:
# Description
Cutouts from compressed files often result in a `BufferUnderflowException`.  This is caused by expecting more bytes than is actually available.

# Changes
**CompressedImageTiler**
- Properly read headers in rather than try to adapt to specific algorithms during decompression (see `initCompressionOption` method).
- Remove reverse traversal from row read in favour of ZNAXIS reverse read

Added tests.


attipaci:
Thanks for the fix, @at88mph!

at88mph:
Cheers, @attipaci .  It occurs to me that I completely forgot to open an Issue.  Do you want that for the Release Notes?

attipaci:
No worries. I just referenced the PR number in the `CHANGELOG.md`, which is what I use for the release notes. So all is good! Thanks again for the fixes!

PR #879, opened 2026-08-05
Fix corrupted pixels when a GZIP_1 tile spans several inflate chunks

timj:
When reading some of the new LSST Data Preview 2 data we noticed:

Screen Shot to Yoink2026-08-05 at 3 29 48 PM

There are weird 150 x ~30 pixel tiles with +/- 1E38 pixels. Our tile sizes are 150 x 150 pixels. The problem is with the tiles that Astropy decides can't be RICE_1 compressed and falls back to using GZIP_1 compression instead. GZIP_2 decompression is fine.

I asked Claude to look at a fix. It definitely seems to fix the problem. Your choice as to whether you want to use the code. Claude also reported the following analysis.

# Fix corrupted pixels when a GZIP_1 tile spans more than one inflate chunk

## Summary

`GZipCompressor.decompress()` assumed that every `GZIPInputStream.read()` returns a whole number
of primitive elements. It does not. Once a tile's decompressed size exceeds the 64 kB internal
buffer, the reads do end in the middle of an element, and the trailing bytes were silently
discarded — so every element after the first chunk was assembled from misaligned bytes.

For a float image this produces values of order 1e38, which destroys the image scaling for any
client that computes a min/max. This was found because the Firefly display tool renders LSST
coadd images unreadably, while astropy and SAOImageDS9 read the same files correctly.

The bug has been present since the original GZIP implementation (1.14.0, 2015).

## The bug

```java
while ((count = zip.read(buffer)) >= 0) {
    if (typeConverter != null) {
        count = typeConverter.copy(count);
    }
    nioBuffer.position(0);
    nioBuffer.limit(count / primitiveSize);   // <-- count % primitiveSize bytes are dropped
    setPixel(pixelData, null);
}
```

`GZIPInputStream.read()` returns whatever the inflater currently has available, which is not
constrained to a multiple of the element size. `count / primitiveSize` truncates, the remaining
`count % primitiveSize` bytes are dropped, and the next iteration restarts at `buffer[0]`. Every
element after the first chunk is therefore built from bytes shifted by however much was dropped,
and the tile's last element is never written at all.

A concrete example, from a 150x150 float tile (90000 bytes) in one of the affected files. The
three reads return 65536, 4919 and 19545 bytes; the second and third are not multiples of 4, so
4 bytes are dropped in total and 22499 of 22500 pixels are written:

| tile row | read sizes | 65536/4 + (2nd read)/4 | first corrupted pixel |
|---:|---|---:|---:|
| 208 | 65536, 4919, 19545 | 16384 + 1229 = 17613 | 17613 |
| 211 | 65536, 10007, 14457 | 16384 + 2501 = 18885 | 18885 |
| 232 | 65536, 8351, 16113 | 16384 + 2087 = 18471 | 18471 |
| 233 | 65536, 15377, 9087 | 16384 + 3844 = 20228 | 20228 |
| 234 | 65536, 9747, 14717 | 16384 + 2436 = 18820 | 18820 |

Each tile's first 16384 pixels are correct and each goes wrong exactly at its own chunk boundary,
confirming the fault is internal to a single tile's stream. Tiles are decoded with independent
`GZipCompressor` instances (`CompressorProvider.TileCompressorControl.decompress()` calls
`newCompressor(option).decompress(in, out)`), so nothing leaks between tiles.

## Which data is affected

Three conditions must hold together:

1. The tile is decoded by GZIP_1. That happens either when `ZCMPTYPE = 'GZIP_1'`, or for any
   individual tile that took the lossless fallback — `getGzipCompressorControl()` hardcodes
   `ZCMPTYPE_GZIP_1` regardless of the primary algorithm.
2. The tile's uncompressed size exceeds `DEFAULT_GZIP_BUFFER_SIZE` (65536 bytes), i.e. more than
   16384 pixels for a 4-byte type.
3. The tile's *compressed* size also exceeds 65536 bytes, so the inflater needs more than one
   input fill and returns a short middle chunk. Highly compressible tiles are delivered in
   aligned chunks and escape unharmed.

`ZCMPTYPE = 'GZIP_2'` is unaffected: `GZip2Compressor.decompress()` accumulates into a full-size
byte array with offset tracking and was always correct.

Condition 1 is easy to hit in practice without any GZIP keyword appearing in the header. cfitsio
(and astropy, which reimplements it) stores an individual tile losslessly whenever quantization
fails for that tile — `fits_quantize_float` returning `status == 0`, surfaced in astropy as
`QuantizationFailedException`, after which the tile is compressed with GZIP_1 and its
`COMPRESSED_DATA` entry is left empty. So a header saying `ZCMPTYPE = 'RICE_1'` can still contain
GZIP_1 tiles.

The LSST/Rubin coadd files that exposed this are `RICE_1` + `SUBTRACTIVE_DITHER_2` with 150x150
float32 tiles (90000 bytes uncompressed). Quantization failed for a minority of tiles for all
three of the usual reasons — constant tiles, tiles containing non-finite values, and tiles whose
required number of quantization levels exceeds int32:

| HDU | fallback tiles | corrupted before this fix | corrupted pixels |
|---|---:|---:|---:|
| IMAGE | 5 of 484 | 5 | 18483 |
| VARIANCE | 222 of 484 | 79 | 205784 |
| NOISE_REALIZATIONS/0 | 116 of 484 | 0 | 0 |
| MASK (GZIP_2) | n/a | 0 | 0 |

Whole-image statistics for the IMAGE HDU of that file, before and after:

| reader | min | max | mean | stddev | NaN |
|---|---|---|---|---|---|
| astropy 8.0.1 | -1675.0472 | 94325.719 | 268.23214 | 1421.4871 | 0 |
| nom-tam-fits, before | -2.6238775e+38 | 2.6136446e+38 | -2.0194821e+32 | 7.5617686e+35 | 36 |
| nom-tam-fits, after | -1675.0472 | 94325.719 | 268.23214 | 1421.4871 | 0 |

Worth noting for anyone assessing impact: a second file from the same dataset, whose IMAGE HDU
contains no fallback tiles and therefore read correctly, still had one corrupted tile (3076
pixels) in its VARIANCE HDU. Because the corruption is per tile and the affected planes are often
not the displayed one, it can be present without being visible.

## The fix

Carry the bytes of a partial element over to the next read instead of discarding them:

```java
int elementSize = typeConverter == null ? primitiveSize : typeConverter.fromSize();

try (GZIPInputStream zip = createGZipInputStream(compressed)) {
    int pending = 0;
    int count;
    while ((count = zip.read(buffer, pending, buffer.length - pending)) >= 0) {
        int available = pending + count;
        int whole = available - available % elementSize;
        if (whole > 0) {
            int converted = typeConverter == null ? whole : typeConverter.copy(whole);
            nioBuffer.position(0);
            nioBuffer.limit(converted / primitiveSize);
            setPixel(pixelData, null);
        }
        pending = available - whole;
        if (pending > 0) {
            System.arraycopy(buffer, whole, buffer, 0, pending);
        }
    }
}
```

Alignment is on the element size *of the compressed stream*, which is `primitiveSize` unless a
`TypeConversion` is active — hence the new `TypeConversion.fromSize()`. Reading into
`buffer[pending..]` keeps the amount handed to `setPixel()` within `nioBuffer`'s capacity, so the
existing conversion arithmetic is untouched.

## Tests

`GZipCompressTest` gains seven tests. Six wrap the gzip stream so that no read returns more than
1023 bytes — not a multiple of 2, 4 or 8 — covering byte, short, int, float, long and double; the
seventh is a plain round trip larger than the internal buffer. Before the fix six of the seven
fail; the byte case correctly passes, since single-byte elements cannot misalign.

`ReadWriteProvidedCompressedImageTest.writeRiceFloatWithForceNoLossTileLargerThanGzipBuffer`
covers the path end to end: a 300x300 float image compressed with `RICE_1` and
`SUBTRACTIVE_DITHER_2` in 300x150 tiles (180000 bytes each), with `forceNoLoss` over the whole
image so both tiles take the GZIP fallback, written out and read back with exact-equality
assertions. Before the fix it fails with `expected: <311.27942> but was: <-2.7092922E-22>`.

Two notes on why that test is built the way it is, since both are easy to get wrong:

- The pixel values are noise, which gzip compresses only about 1.11:1. That ratio is what makes
  the inflater deliver the tile in chunks of 65536, 7449, 65536, 7509, 33970 — the misaligned
  pattern seen in the real files. Written first against the existing `m13real` test image, the
  test passed even without the fix: that data compresses roughly 2:1, so the chunks land at
  65536, 65536, 48928, all multiples of 4.
- The existing `writeRiceFloatWithForceNoLoss` test does exercise the GZIP fallback, but with
  300x15 tiles of 18000 bytes, comfortably inside the 65536-byte buffer. That is why the path was
  covered yet the bug was not caught.

Full suite: 1854 tests, 0 failures, 8 skipped (unchanged).

## Verification against real data

With the fix, both LSST files read identically to astropy 8.0.1 for every float image HDU, and the
five formerly corrupt tiles decode bit-for-bit identically to a plain `GZIPInputStream` gunzip of
the same `GZIP_COMPRESSED_DATA` blob.

One unrelated difference remains and is not addressed here: in the *quantized* tiles the two
readers disagree on many pixels by at most 0.004 absolute, against ZSCALE quantization steps of
0.24 to 6.6 — around 0.1% of one quantization level. This appears equally in files with no GZIP
tiles at all and looks like ordinary rounding in the dequantization arithmetic.

### DIFF
--- a/src/main/java/nom/tam/fits/compression/algorithm/gzip/GZipCompressor.java
+++ b/src/main/java/nom/tam/fits/compression/algorithm/gzip/GZipCompressor.java
@@ -211,11 +211,6 @@ public abstract class GZipCompressor<T extends Buffer> implements ICompressor<T>
             to.putArray(toBuffer, toArray);
             return byteCount * to.size() / from.size();
         }
-
-        /** The size in bytes of one element as stored in the compressed stream. */
-        int fromSize() {
-            return from.size();
-        }
     }
 
     private static final int DEFAULT_GZIP_BUFFER_SIZE = 65536;
@@ -263,26 +258,27 @@ public abstract class GZipCompressor<T extends Buffer> implements ICompressor<T>
 
         // The size of one element as it appears in the compressed stream, which is the element size of the
         // pixel data unless the stream stores a narrower or wider type that needs converting.
-        int elementSize = typeConverter == null ? primitiveSize : typeConverter.fromSize();
+        int elementSize = typeConverter == null ? primitiveSize : typeConverter.from.size();
 
         try (GZIPInputStream zip = createGZipInputStream(compressed)) {
-            int pending = 0;
-            int count;
-            // GZIPInputStream.read() returns whatever the inflater has available, so a chunk may end in the
-            // middle of an element. Only whole elements can be handed on; the trailing bytes of a partial
-            // element are moved to the front of the buffer and completed by the next read.
-            while ((count = zip.read(buffer, pending, buffer.length - pending)) >= 0) {
-                int available = pending + count;
-                int whole = available - available % elementSize;
-                if (whole > 0) {
-                    int converted = typeConverter == null ? whole : typeConverter.copy(whole);
-                    nioBuffer.position(0);
-                    nioBuffer.limit(converted / primitiveSize);
-                    setPixel(pixelData, null);
-                }
-                pending = available - whole;
-                if (pending > 0) {
-                    System.arraycopy(buffer, whole, buffer, 0, pending);
+            int len = 0; // Number of bytes loaded into buffer
+            int nRead = 0; // Number of bytes read from input
+
+            while ((nRead = zip.read(buffer, len, buffer.length - len)) > 0) {
+                len += nRead;
+
+                // Convert only whole elements in buffer
+                int elementBytes = len - len % elementSize;
+                int converted = typeConverter == null ? elementBytes : typeConverter.copy(elementBytes);
+
+                nioBuffer.position(0);
+                nioBuffer.limit(converted / primitiveSize);
+                setPixel(pixelData, null);
+
+                len -= elementBytes; // unprocessed remainder bytes that were already loaded.
+                if (len > 0) {
+                    // Move unprocessed loaded remainder to front of the buffer
+                    System.arraycopy(buffer, elementBytes, buffer, 0, len);
                 }
             }
         } catch (IOException e) {
--- a/src/main/java/nom/tam/fits/compression/algorithm/gzip/GZipCompressor.java
+++ b/src/main/java/nom/tam/fits/compression/algorithm/gzip/GZipCompressor.java
@@ -211,6 +211,11 @@ public abstract class GZipCompressor<T extends Buffer> implements ICompressor<T>
             to.putArray(toBuffer, toArray);
             return byteCount * to.size() / from.size();
         }
+
+        /** The size in bytes of one element as stored in the compressed stream. */
+        int fromSize() {
+            return from.size();
+        }
     }
 
     private static final int DEFAULT_GZIP_BUFFER_SIZE = 65536;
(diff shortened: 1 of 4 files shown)


attipaci:
Thanks @timj,

Thanks for the fix. I'll look through it and digest it -- mainly to make sure I understand what is the issue and how the fix addresses it exactly (I don't trust AI blindly).

The compression implementation in nom-tam-fits predates my involvement, and I really dislike how it's been done -- it is overly complicated, very convoluted, and difficult to follow. Nonetheless, it is now baked in. Ever since I took over (with v1.16) more bugs keep popping up, and I try to fix them as they come. It's my nightmare, but I hope eventually we'll run out of these annoying bugs...

I assume you'd want another release sooner than the targeted Dec 15 date for the next update. Given that 1.22.2 just went out (ahead of its schedule) for the same reason, I'd want to wait a little so we don't produce a release every time a new issue pops up. Let's try to get all LSST related bugs sorted out before the next update. How does September 15 sound like a target date for 1.22.3? 

-- A.

timj:
Thanks. I will discuss timelines with the team but mid-September sounds great -- sorry to be reporting lots of problems. The good thing is that all these GZIP_1 decompression errors occur in areas of the image that are likely to be masked so we can live with it for a short while (it really messes up image scaling in the display). I think @robyww uses Maven for making releases so I'm not sure if he can use a standalone jar file we make ourselves -- will ask.

If you are based at a US institution you do have data rights so you can make an account at https://data.lsst.cloud and we can point you at the full files directly if you want an external verification.

cc/ @gpdf

PR #883, opened 2026-08-14
TableHDU.getColumnMeta() for non-string descriptors and standard keys

### LINKED ISSUE #882: `TableHDU.getColumnMeta()` for non-string column descriptors
We (the Firefly team) found what I think is a bug in `TableHDU.getColumnMeta()`. It returns a `null` in certain cases when it determines that the header value is a number and not a `String`.

See the example header below. My guess is that this has not been found because `TZERO#` is rarely used.

In this case it returns `null`with a call such as `tableHDU.getColumnMeta(2, “TZERO”);`

Below is our work-around code based on how I think it should work. I am not proposing this as a patch. It is just to help you understand.

```java
private static String getColumnMeta(TableHDU<?> hduTable, int icol, String base) {
     Header header= hduTable.getHeader();
     var key= base+(icol+1);
     if (header == null || !header.containsKey(key)) return null;
     var card= hduTable.getHeader().getCard(key);
     return card.getValue();
}
```

___

Image

___

For your reference here is the Firefly PR with  the work-around. https://github.com/Caltech-IPAC/firefly/pull/1994

As an aside... we found this with a Rubin table. They are certainly been doing double-duty in working out our code!

attipaci:
### Added

  - `String TableHDU.getColumnMeta(int index0, IFitsHeader keyBase)` to retrieve column descriptors using the unindexed base form of standard indexable FITS keys.
  
### Changed

 - `String TableHDU.getColumnMeta(int index0, String keyStrem)` to retrieve non-string column data also.
 - `IFitsHeader.n()` changed to throw the more specific `HeaderCardException` instead of the original `IllegalStateException` when the resulting indexed keyword is too long to be a standard FITS keyword. The change is backward compatible. The same applies downstream, e.g. for `TableHDU.setColumnMeta()` / `.getColumnMeta()`.
 - `TableHDU.getColumnName()` no longer trims the string value stored in by the `TTYPEn` keyword, in line with the leading spaces being significant in the FITS standard. The caller may trim the result as necessary when it's not `null`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,9 +13,20 @@ Upcoming bug fix release, likely around 15 September 2026.
 
  - [#879] Fixed GZIP HDU decompression so it properly handles unprocessed remainder data in the internal buffer between successive reads. (by @timj and @attipaci).
 
+
+### Added
+
+ - [883] Added `TableHDU.getColumnMeta(int, IFitsHeader)` to support standard keyword enums beside the existing string keyword form. (by @attipaci)
+
 ### Changed
 
  - [#879] Simplified GZIP2 decompression code. (by @attipaci)
+ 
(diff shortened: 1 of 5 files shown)


robyww:
Looks good. Thanks!

PR #885, opened 2026-08-17
Compression quantization fixes and improvements

### LINKED ISSUE #884: `ZBLANK` value from header is not used in decompression
We have found another (non-rubin) compression issue. This was found by @stargaser while working on the the NEO Surveyor project. He included a small file to demonstrate the problem. The following is an slightly edited version of his Firefly ticket.

___

Overview

When Firefly displays a quantized, tile-compressed floating-point FITS image (ZCMPTYPE = RICE_1 with ZQUANTIZ = SUBTRACTIVE_DITHER_1), pixels that were NaN in the original image are shown as large negative numbers instead of blank/undefined. The image is not corrupt — the file follows the FITS tiled-image-compression convention, and CFITSIO, astropy, and DS9 all reconstruct the NaNs correctly. Decompression is not applying the ZBLANK mapping on read.

Root cause

In the tiled-compression convention, NaNs in a quantized floating-point image cannot be represented in the integer-quantized tile, so they are stored as a reserved integer sentinel given by the ZBLANK keyword (here ZBLANK = -2147483648). A conforming reader must map pixels equal to ZBLANK back to NaN after de-quantization. The decompression appears to d-quantize the sentinel directly (value ≈ ZSCALE × ZBLANK + ZZERO), which yields the large negative values seen. (ZBLANK may appear either as a header keyword or as a per-tile ZBLANK column — a robust fix should handle both.)

Test file — attachment firefly_zblank_nan_test.fits (a 34 KB, project-agnostic repro). It holds the same 64×64 float image (a noisy gradient with a 16×16 NaN block plus a few scattered NaNs) two ways:

EXT 1 ORIGINAL — uncompressed ImageHDU (ground truth).
EXT 2 RICE_QUANT — RICE_1 + SUBTRACTIVE_DITHER_1, ZBLANK = -2147483648.
NEO Surveyor is considering RICE compression and needs the NaNs to render correctly.

___

References

FITS Tiled Image Compression convention (ZQUANTIZ, ZBLANK); Pence, White & Seaman 2010, PASP 122, 1065.
Verified against CFITSIO, astropy, and DS9 (all correct).

___

[firefly_zblank_nan_test.fits.zip](https://github.com/user-attachments/files/31079600/firefly_zblank_nan_test.fits.zip)

### LINKED ISSUE #887: Bug in PLIO decompression of 32-bit integer data
The PLIO decompression of 32-bit integer data did not write the upper 2 bytes due to inappropriate masking. Will Fix in PR #885.

Fortunately PLIO is usually used for compressing pixel masks only (or at least that's what it was designed for), so it typically compresses only 0 or 1 values -- regardless of the integer type used.

attipaci:
### Fixed

 - Fixed header `ZBLANK` value not being used in decompression, when the compressed FITS does not provide per-tile blanking values in a column otherwise. (thanks to @robyww)
 
 - `QuantizeProcessor` had extraneous rounding by half, and sometimes in the wrong direction. Fixed to conform to the FITS specification more completely.
 
 -  Fixed PLIO decompression of 32-bit integer data to restore the upper 2 bytes also.

### Added

 - Added `setColumnData(Object)`, `createColumnData(int)`, and `setColumnSize(int)` methods to `ICompressColumnParameter`.

 - Added `CompressParameters.activeHeaderParameters()` / `.activeColumnParameters()` to selectively return only those parameters that are necessary for describing the tile compression.

 -  `QuantizeOption.useFMA()` method can select whether `Math.fma()` should be used instead of regular arithmetics (default) when converting quantized integers back to their floating-point values. The use of `fma()` matches cfistio / funpack and astropy more closely, but may be very slow on platforms without hardware support. (thanks to @keastrid)

### Changed

 - Quantization in HDU compression has been overhauled, and much simplified. Many strenuous internal classes have been eliminated in favor of simpler, easier to follow, logic.

-  `QuantizeProcessor` integer min/max ranges to use floor / ceil to accommodate dither.
 
 - `QuantizeProcessor` defaults to `ZSCALE` 1.0 and `ZZERO` 0.0 if there are no quantizable valid data (only null or explicit zero values).

 - `Quantize.calculateNoise()` changed to use only regular data, excluding special values like NaNs, infinities, null and zero indicators.
 
 - `Quantize.quantize()` changed to set default values (`ZSCALE` -> 1.0, `ZZERO` -> 0.0, int min/max -> 0) when there is no valid data to quantize.

 - Fully disentangled initialization of compression quantization column data for compression (where prior data may not exist) and decompression (where existing data must be provided).
 
 - Quantized compression to either add `ZBLANK` header value or per-tile column data, but not both, and possibly neither -- as necessary.

 - The default null-value indicator in compressed HDUs is changed to -2147483648, as recommended by the FITS standard.
 
 - Removed `null` options checks from compression parameter methods, and instead commented on the constructors that the parameters should not be instantiated with `null` option parameter. Under the normal behavior of the library the parameters are always created with valid option arguments, so it would take a deliberate effort by the user to do otherwise.

 - Speed up compression / decompression by eliminating extraneous arrays / buffers from the processing. (thanks to @keastrid)

 - `QuantizeOption.toDouble()` method now uses `Math.fma()` to match cfistio / funpack more closely. (thanks to @keastrid)

### Deprecated

 - Deprecated `ICompressColumnParameter.setColumnData(Object, int)`. Its dual functionality has been split into separate `.setColumnData(Object)`, `createColumnData(int)` and `ensureColumnData(int)` methods.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,9 +21,11 @@ Upcoming feature release, likely around 15 September 2026.
 
  - [#883] Added `TableHDU.getColumnMeta(int, IFitsHeader)` to support standard keyword enums beside the existing string keyword form. (by @attipaci)
  
- - [#884] Added `setColumnData(Object)`, `createColumnData(int), and `setColumnSize(int)` methods to `ICompressColumnParameter`. (by @attipaci)
+ - [#884] Added `setColumnData(Object)`, `createColumnData(int)`, and `setColumnSize(int)` methods to `ICompressColumnParameter`. (by @attipaci)
  
  - [#884] Added `CompressParameters.activeHeaderParameters()` / `.activeColumnParameters()` to selectively return only those parameters that are necessary for describing the tile compression. (by @attipaci)
+ 
+ - [#885] Added `QuantizeOption.toInt(double)` and `.toDouble(int)` methods which actually perform the conversion. The `.toDouble()` method now uses `Math.fma()` to match cfistion / funpack more closely (thanks to @keastrid). (by @attipaci)
 
 ### Changed
 
@@ -52,6 +54,10 @@ Upcoming feature release, likely around 15 September 2026.
  - [#885] The default null-value indicator in compressed HDUs is changed to -2147483648, as recommended by the FITS standard. (by @attipaci)
  
  - [#885] Removed `null` options checks from compression parameter methods, and instead commented on the constructors that the parameters should not be instantiated with `null` option parameter. Under the normal behavior of the library the parameters are always created with valid option arguments, so it would take a deliberate effort by the user to do otherwise. (by @attipaci)
+ 
+ - [#885] Speed up compression / decompression by eliminating extraneous arrays / buffers from the processing. (by @attipaci, thanks to @keastrid)
+ 
+ - [#885] `QuantizeOption.toDouble()` method now uses `Math.fma()` to match cfistio / funpack more closely. (by @attipaci, thanks to @keastrid)
 
  - The latest build and runtime Maven dependencies. (by @attipaci)
  
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,8 @@ Upcoming feature release, likely around 15 September 2026.
  - [#884] Fixed header `ZBLANK` value not being used in decompression, when the compressed FITS does not provide per-tile blanking values in a column otherwise. (by @attipaci, thanks to @robyww).
  
  - [#885] `QuantizeProcessor` had extraneous rounding by half, and sometimes in the wrong direction. Fixed to conform to the FITS specification more completely. (by @attipaci)
+ 
+ - [#887] Fixed PLIO decompression of 32-bit integer data to restore the upper 2 bytes also.
 
 ### Added
 
@@ -26,8 +28,6 @@ Upcoming feature release, likely around 15 September 2026.
  - [#884] Added `CompressParameters.activeHeaderParameters()` / `.activeColumnParameters()` to selectively return only those parameters that are necessary for describing the tile compression. (by @attipaci)
  
  - [#885] Added `QuantizeOption.toInt(double)` and `.toDouble(int)` methods which actually perform the conversion. The `.toDouble()` method now uses `Math.fma()` to match cfistion / funpack more closely (thanks to @keastrid). (by @attipaci)
- 
- - [#887] Fixed PLIO decompression of 32-bit integer data to restore the upper 2 bytes also.
 
 ### Changed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,17 +15,15 @@ Upcoming feature release, likely around 15 September 2026.
  
  - [#884] Fixed header `ZBLANK` value not being used in decompression, when the compressed FITS does not provide per-tile blanking values in a column otherwise. (by @attipaci, thanks to @robyww).
  
- - [#884] `QuantizeProcessor` had extraneous rounding by half, and sometimes in the wrong direction. Fixed to conform to the FITS specification more completely. (by @attipaci)
- 
- - [#884] `QuantizeProcessor` integer min/max ranges to use floor / ceil to accomodate dither. (by @attipaci)
+ - [#885] `QuantizeProcessor` had extraneous rounding by half, and sometimes in the wrong direction. Fixed to conform to the FITS specification more completely. (by @attipaci)
 
 ### Added
 
  - [#883] Added `TableHDU.getColumnMeta(int, IFitsHeader)` to support standard keyword enums beside the existing string keyword form. (by @attipaci)
  
- - [#884] Added `setColumnData(Object)`, `createColumnData(int), and `ensureColumnData(int)` methods to `ICompressColumnParameter`. (by @attipaci)
+ - [#884] Added `setColumnData(Object)`, `createColumnData(int), and `setColumnSize(int)` methods to `ICompressColumnParameter`. (by @attipaci)
  
- - [#884] Added `CompressParameters.activeHeaderParameters()` / `.activeColumnParameters()` to selectively return only those parameters that are necessary for describing the tile compression.
+ - [#884] Added `CompressParameters.activeHeaderParameters()` / `.activeColumnParameters()` to selectively return only those parameters that are necessary for describing the tile compression. (by @attipaci)
 
 ### Changed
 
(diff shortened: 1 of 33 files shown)


keastrid:
I was going to put together a PR about this, but since this system is being changed I'll mention it here (I can still make a PR/issue if you prefer). This is based on data from before this PR, I can do another round of profiling with the PR if you think it is ready.

I was recently doing some profiling of fits image opening, and I noticed that `QuantizeProcessor.FloatQuantCompressor#decompress` was taking a while. I found that adding an overload that takes a `FloatBuffer` that does the casting to float performed much better than creating a `DoubleBuffer` then looping through it a second time to convert to floats, e.g.:
```java
public void unquantize(final IntBuffer intData, final FloatBuffer fdata) {
        while (fdata.hasRemaining()) {
            fdata.put((float) pixelFilter.toDouble(intData.get()));
            pixelFilter.nextPixel();
        }
    }
```



There is also a [potential difference](https://redirect.github.com/heasarc/cfitsio/issues/178) in how funpack behaves depending on the system it is built for/enabled optimizations based on the Clang compiler (default for mac) automatically converting the unquantization `d * bScale + bZero` into `fma(d, bScale, bZero)`, which can give a different result depending on the numbers used (should be identical for bScale = 1 and bZero = 0). I am waiting to hear back from heasarc about this. I don't know if you are testing against funpack output, but thought it would be relevant to bring up - though realistically, I don't think any images would be effected outside of some contrived examples. For AIJ, we don't find the different between fma and normal math to be significant, so were going to provide an option to use it since (when it works) it can be significantly faster (though it is proving annoying to profile, usually only marginally faster or the same). We can upstream this if desired.

attipaci:
Thanks @keastrid, 

I'll include your change to the PR.

As for funpack (a.k.a. cfitsio), I don't trust it much. It has many strange quirks and some really weird behaviors. The compression stuff in there even more poorly implemented than in nom-tam-fits (and that's saying something!). In many cases what it does clearly deviates from what is described in the standard -- which is funny given that both are by the same people. The bottom line is that it's not a very good reference implementation. I think astropy may be a better one to compare to. They also found the cfitsio implementation not great, and decided to build it from scratch.

attipaci:
Hi @keastrid, 

Thanks again for the insight. I took your suggestion and I've made it into a slightly bigger overhaul, eliminating other internal arrays / buffers also -- thus further streamlining the quantized compression / decompression of both floats and doubles. Let me know if there is more to be done while at it...

keastrid:
Hi Attilla,

Thanks, this seems to work well on my end!

Checking astropy's mac build, it does seem to use fma in the compiled result. For nom.tam, it might be better to have a fallback in case the system does not support fma as the java implementation it falls back to is very slow. Having a static final field with the result of a system property should allow the JIT to eliminate the conditional entirely so shouldn't affect performance. The systems in question would be those older than [~2012](https://en.wikipedia.org/wiki/FMA_instruction_set#FMA3_instruction_set), most arm computers after 2011 should support it.

I don't see other areas worth pursuing, unless you have any ideas on speeding up the IO or decompression itself.

attipaci:
Hi @keastrid,

I'm not sure it's worth complicating the `fma()` part for architectures that are 15+ years old. So, I vote for just keeping it simple, even if it means a performance hit on outdated platforms.

keastrid:
Hi Attila,

I agree that it is unlikely to encounter running nom.tam on hardware that lacks support for it. Whether to give it a toggle or not is up to you; however, I think you are underestimating just how slow the fallback is. You can disable hardware fma with the launch parameter `-XX:-UseFMA` if you'd like to test with it.

attipaci:
Hi @keastrid!

Well, if the performance penalty is really so severe, I think we have the following options:

 1. Use a system property as you suggest. What system property you have in mind, exactly? And would that be a foolproof and future-proof way (e.g. with possibly emerging new architectures) to detect when the architecture does not support `fma()` in hardware?
 2. Add a static method to allow the user to toggle between `fma()` and basic arithmetic at runtime.
 3. Revert back to the old arithmetic. This is what the FITS sandard describes, with no mention of `fma()`. Any discrepancy between `fma()` and the arithmetic method will likely be totally insignificant (i.e. orders of magnitude below the quantization level) in every realistic use case.

Given the imperfect choices, I'm leaning towards option 3 as the most robust one -- but I can be convinced of one of the others if you give me a solid pitch.

keastrid:
Hi Attila,

I think 1 is the best of both worlds. I was going for something along the lines of `nom.tam.fits.useFma/disableFma` (whichever you prefer to be the default). 

The JVM will automatically set `UseFMA` for itself on supported hardware, so we can reuse that for this.
```java
// Automatically enable `nom.tam.fits.useFma` system property if JVM UseFMA is enabled.
// The JVM will automatically enable the UseFMA vm option if the cpu supports it.
// Not available on openj9
try {
    HotSpotDiagnosticMXBean hotSpotDiagnostic = ManagementFactory.getPlatformMXBean(HotSpotDiagnosticMXBean.class);
    VMOption useFMAVMOption = hotSpotDiagnostic.getVMOption("UseFMA");
    boolean useFMA = Boolean.parseBoolean(useFMAVMOption.getValue());
    if (useFMA) {
        System.setProperty("nom.tam.fits.useFma", String.valueOf(true));
    }
} catch (Throwable t) {
    // Likely on a different runtime
}
```
A version avoiding the compile time dep in case you prefer this approach:
```java
// Automatically enable `nom.tam.fits.useFma` system property if JVM UseFMA is enabled.
// The JVM will automatically enable the UseFMA vm option if the cpu supports it.
// Not available on openj9
try {
    MBeanServer server = ManagementFactory.getPlatformMBeanServer();
    ObjectName name = new ObjectName("com.sun.management:type=HotSpotDiagnostic");

    if (server.isRegistered(name) && server.invoke(name, "getVMOption", new Object[]{"UseFMA"}, new String[]{"java.lang.String"}) instanceof CompositeData compositeData) {
        var useFma = Boolean.parseBoolean(compositeData.get("value").toString());
        System.setProperty("nom.tam.fits.useFma", Boolean.toString(useFma));
    }
} catch (ReflectionException | MBeanException e) {
    // Failed to execute MBean operation
} catch (MalformedObjectNameException _) {

} catch (InstanceNotFoundException _) {
    // Likely not on HotSpot
}
```

attipaci:
Hi @keastrid,

My issue with your solution is that it not guaranteed to work on non-Hotspot VMs (as noted in the `InstanceNotFoundException` catch). So, it does not provide a universal solution. I'll revert back to the good old arithmetic...

keastrid:
Alright, it's a patch we can maintain easily enough.

attipaci:
Hi @keastrid,

To make your life a little easier, I added a `QuantizeOption.useFMA(boolean)` static method to let you select `fma()` at runtime. (Regular arithmetic remains default as it is more universal.)

So, instead of having to maintain a patch, you can just add your static detector snippet into your application to call that method. E.g.,

```java
// Use fma() if JVM UseFMA is enabled.
// The JVM will automatically enable the UseFMA vm option if the cpu supports it.
// Not available on openj9
try {
    MBeanServer server = ManagementFactory.getPlatformMBeanServer();
    ObjectName name = new ObjectName("com.sun.management:type=HotSpotDiagnostic");

    if (server.isRegistered(name) && server.invoke(name, "getVMOption", new Object[]{"UseFMA"}, new String[]{"java.lang.String"}) instanceof CompositeData compositeData) {
        QuantizeOption.useFMA(Boolean.parseBoolean(compositeData.get("value")));
    }
} catch (Exception e) {
    QuantizeOption.useFMA(false);
}
```

keastrid:
Hi Atilla, 

That's perfect, thanks!

PR #891, opened 2026-08-30
Fixes to `SUBTRACTIVE_DITHER_2` implementation

### LINKED ISSUE #892: Wrong zero indicator for `SUBTRACTIVE_DITHER_2`
The FITS standard designates -2147483647 as the integer to denote 0.0 values but the library used -2147483646 instead which is wrong.

attipaci:
### Fixed
  
 - [#892] When `SUBTRACTIVE_DITHER_2` was used (via `QuantizeOption.setDither()` and `.setDither2()`), the library used the wrong 0.0 value indicator (-2147483646), instead of the value -2147483647 prescribed by the standard. Fixed by switching to the standard indicator value.

### Changed

- `QuantizeOption.setDither2()` should not force setting `ZZERO` to 0.0 (it was), or enable dithering itself. Instead, it needs only to enable or disable the special treatment of 0.0 values, as per the FITS specification. Fixed to conform to standard more closely, and updated the Javadoc to reflect its proper usage. The old behavior was not particularly troublesome, but it was an unnecessary quirk of the implementation.
 
 - `QuantizeOption.setCenterOnZero()` did not produce the advertised behavior of keeping ZZERO at 0.0. Changed implementation to match the contract of this method. As such when `ZZERO` is not forced to be 0.0, it will be chosen to try quantize with positive integers only, which can make compression more efficient in some cases.
 
 - Automatic quantization for the compression of floating-point data as integers (via `Quantize.quantize()`) returned `false` (failure to quantize) in more cases than needed. Quantized representation is not possible only if the quantized range exceeds the 32-bit integer data range available, exclusing the special values). The change improves the compression of data. 

 - `TableHDU.getColumn(String)` changed to return `null` when there is no column with the specified name. Previously, it threw an `IndexOutOfBoundsException`.

### Deprecated

- Deprecated `QuantizeOption.setCheckZero()` and `.isCheckZero()` methods. They are duplicates of the `setDither2()` and `isDither2()` methods respectively. Similarly, `.setCheckNull()` and `.isCheckNull()` methods are deprecated as checking for NaN values is automatic when integer compressing floating-point values.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,8 +18,6 @@ Upcoming feature release, likely around 15 September 2026. Many fixes and improv
  - [#885] `QuantizeProcessor` had extraneous rounding by half, and sometimes in the wrong direction. Fixed to conform to the FITS specification more completely. (by @attipaci)
  
  - [#887] Fixed PLIO decompression of 32-bit integer data to restore the upper 2 bytes also.
- 
- - [#891] `QuantizeOption.setDither2()` should not enable/disable zero leveling (it was), or dithering itself. Instead, needs only to enable or disable the special treatment of 0.0 values, as per the FITS specification. Fixed to conform to standard more closely, and updated the Javadoc to reflect its proper usage.
   
  - [#892] When `SUBTRACTIVE_DITHER_2` was used (via `QuantizeOption.setDither()` and `.setDither2()`), the library used the wrong 0.0 value indicator (-2147483646), instead of the value -2147483647 desginated by the standard. Fixed by switching to the standard indicator value. (by @attipaci)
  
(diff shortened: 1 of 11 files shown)


attipaci:
Hi @keastrid and @robyww,

This is yet another PR meant to clean up more of the quantized compression mess from before. At last, I think the quantized compression follows what is described in the FITS standard quite closely and completely. Also, the implementation of the related methods now matches their contract (and intention) better. There were some awful discrepancies but most of them harmless in the sense that they did not produce invalid data, only that the compression options set did not always do what they were advertised for. 

Please check it out to see if this PR breaks anything on your end (I should not expect it to). Then next week, I'll plan to create a second rc release, and if no further issues come up, I think we are on target for a Sep 15 release.

Thanks in advance.

PR #894, opened 2026-09-05
Consolidate FitsDate year formatting and extend year range

traitimtrongvag:
Fixed

- "FitsDate" did not correctly handle FITS dates with years outside the four-digit range. Years before 1000 were formatted without the required leading zeroes, and years outside the supported four- or five-digit FITS representation were not handled correctly. Fixed to support years from "-99999" to "99999" using the FITS-specified year formats.

Changed

- "FitsDate.toString()" now uses the same FITS date formatting logic as "getFitsDateString(Date, boolean)", ensuring consistent handling of astronomical years, including negative and extended years.

- Year parsing now supports both signed and unsigned FITS year representations and allows non-canonical signed forms to be parsed and normalized.

- Added and updated tests covering four-digit, extended, negative, boundary, and non-canonical FITS year representations.

### DIFF
--- a/src/main/java/nom/tam/fits/FitsDate.java
+++ b/src/main/java/nom/tam/fits/FitsDate.java
@@ -48,6 +48,8 @@ public class FitsDate implements Comparable<FitsDate> {
      * logger to log to.
      */
 
+    private static final int FIRST_FOUR_CHARACTER_VALUE = 1000;
+
     private static final int FIRST_THREE_CHARACTER_VALUE = 100;
 
     private static final int FIRST_TWO_CHARACTER_VALUE = 10;
@@ -263,7 +265,7 @@ public class FitsDate implements Comparable<FitsDate> {
             return "";
         }
         StringBuilder buf = new StringBuilder(FitsDate.FITS_DATE_STRING_SIZE);
-        buf.append(year);
+        appendFourDigitValue(buf, year);
         buf.append('-');
         appendTwoDigitValue(buf, month);
         buf.append('-');
@@ -334,6 +336,13 @@ public class FitsDate implements Comparable<FitsDate> {
         }
 
         return Integer.compare(millisecond, fitsDate.millisecond);
+    }
+
+    private void appendFourDigitValue(StringBuilder buf, int value) {
+        if (value < FitsDate.FIRST_FOUR_CHARACTER_VALUE) {
+            buf.append('0');
+        }
+        appendThreeDigitValue(buf, value);
     }
 
     private void appendThreeDigitValue(StringBuilder buf, int value) {
--- a/src/main/java/nom/tam/fits/FitsDate.java
+++ b/src/main/java/nom/tam/fits/FitsDate.java
@@ -152,7 +152,7 @@ public class FitsDate implements Comparable<FitsDate> {
 
         int fitsYear = toFitsYear(cal.get(Calendar.ERA), cal.get(Calendar.YEAR));
 
-        StringBuilder fitsDate = new StringBuilder(FitsDate.FITS_DATE_STRING_SIZE);
+        StringBuilder fitsDate = new StringBuilder(FITS_DATE_STRING_SIZE);
         appendYear(fitsDate, fitsYear);
         fitsDate.append('-');
         appendTwoDigitValue(fitsDate, cal.get(Calendar.MONTH) + 1);
@@ -188,17 +188,13 @@ public class FitsDate implements Comparable<FitsDate> {
      * @throws FitsException if the year is outside of the range that FITS can represent.
      */
     private static void appendYear(StringBuilder buf, int year) {
-        if (year < -FitsDate.MAX_FITS_YEAR || year > FitsDate.MAX_FITS_YEAR) {
+        if (year < -MAX_FITS_YEAR || year > MAX_FITS_YEAR) {
             throw new FitsException(
-                    "Year " + year + " is outside of the range [-" + FitsDate.MAX_FITS_YEAR + ":" + FitsDate.MAX_FITS_YEAR
+                    "Year " + year + " is outside of the range [-" + MAX_FITS_YEAR + ":" + MAX_FITS_YEAR
                             + "] that a FITS date can represent");
         }
-        if (year > FitsDate.MAX_FOUR_DIGIT_YEAR) {
-            buf.append('+');
+        if (year < 0 || year > MAX_FOUR_DIGIT_YEAR) {
             appendFiveDigitValue(buf, year);
-        } else if (year < 0) {
-            buf.append('-');
-            appendFiveDigitValue(buf, -year);
         } else {
             appendFourDigitValue(buf, year);
         }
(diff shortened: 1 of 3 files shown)


attipaci:
Thanks @traitimtrongvag,

I marked this as an enhancement, not a bug, because it is unlikely anyone was using the library to write data taken prior to 1000 AD. :-D 

However, if we extends the date range support for years 0-999, we could (should) also consider dates between -999 (= 1000 BC) and  -1 (= 2 BC). And, we should throw a `FitsException` for dates that do not fit into the 4-digit year specification, i.e. &lt; -999 or &gt; 9999.

Would you like to amend the PR with the extra checks, and for handling of negative years in the -999 to -1 range?

traitimtrongvag:
Sounds good, I’ll update the PR accordingly.

attipaci:
PS. I also think that handling the date-to-string conversion in different ways in `getFitsDateString()` and `FitsDate.toString()` is a bad idea because it is divergent. I think the better solution would be to consolidate:

- Fix up `getFitsDateString(Date, boolean)` to handle years -999 to 9999, and throw a `FitxException` for dates outside.
- Change `FitsDate.toString()` to simply call `getFitsDateString(toDate(),  true)`. While it is less comptutationally efficient than the current implementation, it has the advantage that it is guaranteed to be consistent. And, I don't think the performance is an issue, since it is extremely unlikely one would call `FitsDate.toString()` on billions of dates in quick succession.

attipaci:
One more thing,

I just checked the FITS standard (Section 9.1.1). And it allows for 5-digit years with a mandatory sign in front, i.e.:

```
[{+,-}C]CCYY 
```

In fact, negative years always require the 5 digit format.

So years 0 to 9999 can (and _should_) use 4 digits, e.g. 2026. Years 10000 through 99999 _must_ use the signed 5-digit representation, e.g. `+10000` to `+99999`, and similarly negative years _must_ use the 5-digit format, e.g. `-99999` to `-00001`. 

Would you be up to implement all that in `getFitsDateString(Date, boolean)`, and then let `FitsDate.toString()` call that like described above?

traitimtrongvag:
Sure, happy to do that :-) I’ll update it to follow the FITS standard and have "getFitsDateString()" handle the formatting.

attipaci:
Thanks @traitimtrongvag!

src/main/java/nom/tam/fits/FitsDate.java
```diff
@@ -202,6 +256,30 @@ public FitsDate(String dStr) throws FitsException {
...
 
+    /**
+     * Parses the signed/unsigned FITS year token matched by {@link #NORMAL_REGEX}.
+     *
+     * @throws FitsException if the extended year token is out of range for its sign (e.g. <code>+00001</code> or
+     *                           <code>-00000</code>).
+     */
+    private static int parseYear(String yearToken) throws FitsException {
+        char sign = yearToken.charAt(0);
+        if (sign == '+') {
```
attipaci: I think this method can be made much simpler:

```java
  try {
    int year = Integer.parseInt(yearToken);
  } catch (NumberFormatException e) {
    throw(new FitsException("Invalid year specification: " + year);
  }
```

Since `Integer.parseInt()` handles both the signed and unsigned parsing, and we should not care about the number of digits here, since the regex matching in the constructor that calls this function readily checks for bounds. (And even if it does not, we should not care too much. It's OK to be able to parse dates that are not entirely standard.)

src/main/java/nom/tam/fits/FitsDate.java
```diff
@@ -114,42 +129,81 @@ public static String getFitsDateString(Date epoch) {
...
+     *
+     * @throws FitsException if the year is outside of the range that FITS can represent.
+     */
+    private static void appendYear(StringBuilder buf, int year) {
+        if (year < -FitsDate.MAX_FITS_YEAR || year > FitsDate.MAX_FITS_YEAR) {
+            throw new FitsException(
+                    "Year " + year + " is outside of the range [-" + FitsDate.MAX_FITS_YEAR + ":" + FitsDate.MAX_FITS_YEAR
+                            + "] that a FITS date can represent");
+        }
+        if (year > FitsDate.MAX_FOUR_DIGIT_YEAR) {
```
attipaci: I think I would move the sign into `appendFiveDigitYearValue()`, so here it could be:

```java
if (year < 0 || year > FitsDate.MAX_FOUR_DIGIT_YEAR) {
  appendFiveDigitValue(buf, year);
} else ...
```

src/main/java/nom/tam/fits/FitsDate.java
```diff
@@ -336,14 +409,28 @@ public int compareTo(FitsDate fitsDate) {
...
-    private void appendThreeDigitValue(StringBuilder buf, int value) {
+    private static void appendFourDigitValue(StringBuilder buf, int value) {
+        if (value < FitsDate.FIRST_FOUR_CHARACTER_VALUE) {
+            buf.append('0');
+        }
+        appendThreeDigitValue(buf, value);
+    }
+
+    private static void appendFiveDigitValue(StringBuilder buf, int value) {
+        if (value < FitsDate.FIRST_FIVE_CHARACTER_VALUE) {
```
attipaci: So the sign here:

```java
  if (value < 0) {
    buf.append('-');
    value = -value;
  } else {
    buf.append('+');
  }

  if (value < FIRST_FIVE_CHARACTER_VALUE) ...
```

src/main/java/nom/tam/fits/FitsDate.java
```diff
@@ -114,42 +129,81 @@ public static String getFitsDateString(Date epoch) {
...
-        fitsDate.append(df.format(cal.get(Calendar.YEAR)));
-        fitsDate.append("-");
-        df = new DecimalFormat("00");
 
-        fitsDate.append(df.format(cal.get(Calendar.MONTH) + 1));
-        fitsDate.append("-");
-        fitsDate.append(df.format(cal.get(Calendar.DAY_OF_MONTH)));
+        int fitsYear = toFitsYear(cal.get(Calendar.ERA), cal.get(Calendar.YEAR));
+
+        StringBuilder fitsDate = new StringBuilder(FitsDate.FITS_DATE_STRING_SIZE);
```
attipaci: I think adding `FitsDate.` before the private constrants is superfluous. So, I'd change e.g. `FitsDate.FITS_DATE_STRING_SIZE` to simply `FITS_DATE_STRING_SIZE` here and everywhere where private constants are used.

attipaci (changes requested):
Thanks for implementing it. I think its pretty good overall. I just have a  few suggestions, mainly stylistic. If you agree but don't want to spend the time on them, I'll be happy to amend the PR myself, so long as you consent to it. ;-)

traitimtrongvag:
I’ve addressed all the suggested changes. I also simplified "parseYear()", moved the sign handling into "appendFiveDigitValue()", and removed the unnecessary "FitsDate." qualification from private constants.

attipaci:
Wow, that was quick! Thanks again. ;-)

I see that the `parseYear()` exception is not covered. I think it cannot be reached given the regex matching before calling `parseYear()` (i.e., the regex match ensures that `parseYear()` always gets a parseable value). It was my bad suggesting the `try` / `catch` construct without thinking about it more carefully.

So, as a last change you can remove `parseYear()` entirely simply inline the `Integer.parseInt()` in the constructor instead (without `try` / `catch`)...  (Or, I can do that, if you prefer...)

traitimtrongvag:
Okay, got it! I understand, and I’ll do that right now.

traitimtrongvag:
Done! I removed "parseYear()" and inlined "Integer.parseInt()" as suggested. Thanks for catching that! :)

attipaci (approved):

PR #900, opened 2026-09-12
Fix stale option in copied quantize parameters

traitimtrongvag:
Copied QuantizeParameters kept the source QuantizeOption, causing ZBLANK to use the wrong BNull value.

This rebinds the copied parameters to the destination option and adds regression tests for both cases.


attipaci:
Thanks @traitimtrongvag, looks like another good catch!

